PR #2, opened 2017-09-11
Removed duplicate sys-libs/gpm entry

darkbasic:
Great work. It would be interesting to add another package.env file to filter out -O3 optimizations from packages which actually perform worse with -O3.


InBetweenNames:
Thanks!  And I agree with your suggestion.  It would be cool if we could get people to test some of the more complex packages with "O2" and "Graphite-O3" settings to see if there are regressions.  Since we don't override any of the compiler's judgements, these would form the basis for some good GCC bug reports!  We already have a few worthy bug reports for Graphite from this repo that I haven't gotten around to filing yet.

PR #8, opened 2017-09-15
Add app-editors/vim to ltoworkarounds.conf

Althorion:
Vim doesn’t build with LTO with perl support, but builds fine without it.


InBetweenNames:
Thanks for the contribution!

PR #13, opened 2017-09-18
As this repository is in git, we should consider using thin-manifests.

mgomersbach:
This will prevent recomputing and unnecessary collisions.
See also: https://github.com/gentoo-audio/audio-overlay/pull/3#discussion_r117957565 and https://wiki.gentoo.org/wiki/Repository_format/package/Manifest#Thin_Manifest


InBetweenNames:
Yes that's very true, I support this idea

PR #16, opened 2017-09-18
Added support for multiple sets.

mgomersbach:
Moved set file from ebuild to repo. @deep could be source for bin mirror later on.
Added tc (toolchain) set for updated docker images. Official gentoo images use stable toolchain.
With "emerge -uDN @tc" one can update just the toolchain.


InBetweenNames:
The prospect of generating binary packages with a bin mirror is interesting.  And I do like the idea of making it easier to make Docker images.  I have just one suggestion--could we pick some more intuitive names for them?  For example @tc could become @toolchain and @deep could become @ltoknownworking.  With `worldsetdeep` becoming a set, I see this as making it easier to judge coverage of `/usr/portage` as well.

mgomersbach:
Now we just wait for the build :P

InBetweenNames:
Great, thanks for the contribution!

PR #22, opened 2017-09-22
Pkg moarvm made ltofat.conf so nqp and rakudo can build for Perl6.

preney:
The package moarvm (an indirect dependency for Perl6) needs to be ltofat.conf for dev-lang/nqp and dev-lang/rakudo to compile and link.


InBetweenNames:
Thanks!

PR #25, opened 2017-09-23
QEMU requries glib to be ltofat.conf, no-strict-aliasing, and no LTO.

preney:
Unfortunately, compiling QEMU with LTO causes these issues:

- code may be misoptimized unless -fno-strict-aliasing is used
- many type mismatches (e.g., sizeof issues; -Wlto-type-mismatch) with LTO

This pull request sets -fno-strict-aliasing and -fno-lto for QEMU, and, sets glib to be ltofat.conf.


InBetweenNames:
Thanks @preney!

PR #26, opened 2017-09-24
Convert to full profile, seperate optimization techniques. Update build process and ebuild.

mgomersbach:
The overlay is now selectable as a profile!

"eselect profile list" will show our first profile.

For now the O3 and and graphite are not selected by default.
Also I have just created the AMD64 arch profile, maybe later we can extend to others.

I want to first fully test/build world with just LTO (as it's "most difficult")

The splitting of these techniques will make the applying of env vars easier and better debug-able.

The ltoize ebuild has been updated to apply the needed symlinks and the extra files have been moved to the profile it belongs to.
The o3, graphite and lto introduced use flags will make sure the ebuild applies the correct symlinks (and even better, on update: updates, on remove: removes) 

Also the Travis build has now been extended. Before we just had a check on the repo.
But now I've added a second step that we can use to create a binhost!
We've seen some user say he can't use our overlay because he does not have the compute power.
I understand this, because a lot of the stuff we do needs A LOT of compiling.
So with the binhost they would get a head-start (And if they change their mind they only need 1 more compile of world to get on march "native")
The new docker image is one I made recently to also support other overlays. (overlay-docker)

Thanks to Shane, we added our first "base" useflag "pgo"! (as it does not depend on anything or changes stuff wildly like LTO does)

The only bummer I have to report here is that, because the overlay is added to the global list: local testing does not work yet.
Layman needs a patch to support the cli argument "-o" to override syncing all repo's
The result is that 3/4 of my time spent on this pull request was lost to debugging why my custom url just wouldn't want to work and kept using Shane's as the source for it's overlay.


mgomersbach:
So do not hesitate to postpone PR until layman is fixed. (It would be nice though!)

InBetweenNames:
:O holy crap, you did it!  We actually have LTO profiles now!  There are a few merge conflicts to work around but once we get those worked out, I absolutely want this in.  Just a few questions:

* in the current repo we've got separate configs for Graphite and LTO overrides, but the default config is to enable them both with O3.  So we could have `media-video/ffmpeg lto/nolto.conf lto/nographite.conf` which would basically append `-fno-lto` and `-fno-{loop-nest-optimize,graphite-identity,...}` to `{C,CXX,LD}FLAGS`.  Then `/etc/portage/bashrc` runs and removes the redundant flags.  In this PR, it looks like O3, Graphite, and LTO are separate profiles--is it possible to combine them?  I'm thinking it may be better to have just one `lto-overlay` profile and install all overrides directly in that--users who don't use Graphite or O3 will be entirely unaffected that way.  Less maintenance too.

* Can you tell me how on earth you figured this out?  I spent a solid couple of hours trying to figure out how to add env overrides in profiles and turned up absolutely nothing except a couple of wishful forum posts.

All in all awesome work and thank you for the contribution!  I won't be able to merge this today but it will be next up!

mgomersbach:
Hey,
Sorry for the confusion: O3 and graphite are just not linked in the base profile, feel free to add them in profiles/lto-overlay/default/linux/amd64/13.0/parent .

They are seperate so people can use what they want. They are however cumulative: this means nolto.conf can be in LTO and noo3.conf can be in O3. When both selected through the ltoize useflags, these will be combined. This will prevent huge files of workaround and makes it easier to develop.

mgomersbach:
As for the second part: The env vars are not actually from the profile. ltoize is still needed to create symlink from the /var/lib/layman/OVERLAY/etcetera folder.

darkbasic:
Great work, regarding the last commit (the one you did by mistake), you can simply run `git rebase -i HEAD~3` and then change `pick` with `drop` for the last two commits. The you will have to run `git push origin master --force` and the PR will get automatically updated. It's not something you would like to do in a public repo, but it's the perfectly fine in a PR.

mgomersbach:
Thanks :)

I'll go to bed now.

darkbasic:
Why did you close it?

InBetweenNames:
@mgomersbach Just so you know I'm still planning on merging this -- it's just there's a lot of commits to go through and I'll have to devote a chunk of time to sorting it out.  I'm a little time constrained as I'm in a crunch time right now, but rest assured I have not forgotten about it.

mgomersbach:
No worries, I did that to rebase it around the mv ebuild

mgomersbach:
And fix the problems with layman and the CI

Issue #28, opened 2017-09-26
Fix the travis config to pull in the mv overlay


InBetweenNames:
As this overlay now depends on the mv overlay, we'll have to amend the CI tests to pull in that overlay as well

mgomersbach:
We currently just copy the overlay in /usr/local/portage and a piece of local storage in travis (the repo)
This is not very compatible with multiple overlays.

Preferably I'd use layman. It does it cleanly and concisely.

One of the problems I have with layman is dependency:
This [bug](https://bugs.gentoo.org/545568)  , has someone working on it [here](https://github.com/gentoo/layman/compare/master...Hummer12007:recursive) and [here](https://github.com/gentoo/layman/pull/40)

The other is development workflow:
When a repository is listed in the global overlay list, it is IMPOSSIBLE to supply another upstream resource.
The repository from the global list always takes precedence because the URL's from the command-line option are APPENDED. So forks will never be able to use CI.

Sadly the development on layman staggers a bit because they seem to stay compatible with Gentoo derivatives.
Understandable but we need to creative til then...

InBetweenNames:
I wish I knew more about docker and Travis!  I wonder if it may be possible to directly write a "repos.conf" file that contains the "mv" overlay?  For example:

~~~
[mv]
location = /usr/local/mv
sync-type = git
sync-uri = https://anongit.gentoo.org/git/user/mv.git
auto-sync = Yes
~~~

As long as that's done before the `emerge --sync`, it should be automatically pulled in and populated.  Is that possible to do?

mgomersbach:
That might be an idea.
Coincidentally; if we just use that ebuild, should we include the entire overlay of mv?

InBetweenNames:
If we could pull in just the ebuild that actually might work.  Although the `layout.conf` might make repoman complain since it does list the `mv` dependency in there.  Worth a shot I think!

darkbasic:
Can't you just run `layman -a mv` inside the docker container?

mgomersbach:
Yes we can, but should we?
That is one extensive overlay and would like to keep this clean.

InBetweenNames:
I've pushed up a hack to get it to work for now.  When we come up with a better solution we can replace this with it.

InBetweenNames:
Argh.  I'm just going to comment it out until we figure it out.  Unfortunately I can't spend any more time on it today.

darkbasic:
> I've pushed up a hack to get it to work for now.

[...]

PR #29, opened 2017-09-27
Integrate mv ebuild

mgomersbach:
Repoman did not like the mv in masters. Reverted the mv in masters.
Repoman showed the ebuild to be pulling in dependencies. The ebuild has been modified to exclude title features.

This should give us a successful CI for now. And peace of mind as we just need to track upstream changes which will be a lot simpler than a mask/unmask circus and added cruft.


mgomersbach:
#28

Alessandro-Barbieri:
Why repoman doesn't like MV in master?

InBetweenNames:
I suspect because the MV overlay wasn't actually pulled in, so it was complaining that it didn't exist.  Instead of having a separate ebuild, I patched it to do a two-step `sync`.  This way it's less maintenance and we will play nice with existing installations of `app-portage/portage-bashrc-mv`.

PR #31, opened 2017-10-07
Turn off LTO with packages to allow genkernel initramfs w/LVM+LUKS to work.

preney:
LTO has to be turned off for a number of packages so that genkernel can successfully built an initramfs supporting LVM+LUKS (e.g., for an encrypted root filesystem).


preney:
I have tested everything with the removed *FLAGS-=...* (i.e., emerge -1av + genkernel all + a reboot) and everything builds properly and rebooting was successful.

InBetweenNames:
Thanks for the contribution @preney !

PR #44, opened 2017-10-14
More info + packages

pchome:
While I almost finished rebuilding my world w/ LTO (I'll wait a few days for  kde-frameworks 5.39) here is information I collected.

Please note: converted from my package.env so I'm not sure about flag-format in package.cflags.

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -73,7 +73,7 @@ kde-apps/dolphin *FLAGS-=-flto*
 
 # BEGIN: Additional flags required
 #dev-qt/qtcore EXTRA_CXXFLAGS="${CXXFLAGS}" EXTRA_LFLAGS="${LDFLAGS}" # hack to build QtCore w/ LTO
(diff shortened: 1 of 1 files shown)


InBetweenNames:
Thanks! This looks very good.  I just have a couple of questions -- do the `adplug` and `ncurses` errors only manifest under LTO?  Or do they manifest even if a non-LTO build is made?

pchome:
Not sure, but they was installed w/ `-march=native -O2 -pipe` flags. You can check them with `# FLAGS="..." emerge -1 -B ...` command w/o (re)installation.

https://bugs.gentoo.org/580992 and https://bugs.gentoo.org/606142

InBetweenNames:
Just checked them.  The `ncurses` workaround is for sure necessary for LTO, but the `adplug` one actually seems to be independent of LTO -- it manifests under both LTO and non-LTO conditions.  Would you be willing to revise this PR to remove the `adplug` entry from `ltoworkarounds.conf`?  I just want to make sure things are kept specific to LTO.

InBetweenNames:
And by the way, ncurses was a pet peeve of mine for quite a while, so thanks for finding a workaround for that!

InBetweenNames:
Thanks @pchome !

PR #47, opened 2017-10-19
media-video/ffmpeg: can be built w/ LTO

pchome:


InBetweenNames:
Hi @pchome , I'm actually getting some ODR violations when I try to build it with LTO.  What are your CFLAGS/CXXFLAGS for this one?

pchome:
CFLAGS/CXXFLAGS same as yours (from `make.conf.lto`)
```sh
Version: 3.4(0/55.57.57)
USE flags set: X alsa bzip2 encode gpl hardcoded-tables iconv jack ladspa mp3 network openal opengl openssl postproc pulseaudio sdl svg threads truetype vdpau vorbis x264 xcb xvid zlib 
ABI_X86="64 -32 -x32"
CPU_FLAGS_X86="3dnow 3dnowext mmx mmxext sse sse2 sse3 -aes -avx -avx2 -fma3 -fma4 -sse4_1 -sse4_2 -ssse3 -xop"
FFTOOLS="aviocat cws2fws ffescape ffeval ffhash fourcc2pixfmt graph2dot ismindex pktdumper qt-faststart sidxindex trasher"
```

CrisBRM:
@InBetweenNames I can confirm. I actually tried to build ffmpeg with -flto yesterday or two days ago and it failed spectacularly. 
@pchome any ffmpeg benchmarks?

pchome:
Maybe it builds for me because I have every dependency LTOed or have no problematic USE flag set?
Better if someone show build error or attach build.log there ⬇️ in comment block .

@CrisBRM no, nothing to compare with and I don't want to play with it for now.

InBetweenNames:
I actually just did an `emerge -1e ffmpeg` just to see if that was the case, and I still get an error:

~~~
error: `asm` operand has impossible constraints
~~~

CrisBRM:
@pchome if you have mpv you can easily benchmark it. Add this to mpv.conf 
```[bench]
audio=no
untimed=yes
video-sync=display-desync
vulkan-swap-mode=immediate
opengl-swapinterval=0
osd-msg1="FPS: ${estimated-display-fps}"```

... and then, ```mpv --no-config --scale=ewa_lanczossharp --cscale=ewa_lanczossharp --profile=bench /path/to/video```. You'd just have to reemerge ffmpeg without lto...

InBetweenNames:
I'm actually wondering if it is from the `CPU_FLAGS_X86`, since @pchome has a more restricted set than I do.  I could see that resulting in a codegen error with `asm` usages.

InBetweenNames:
Upon further investigation it looks like this is actually related to SSE.  I'll make a small note in the workarounds file to reflect this, as we can't toggle LTO based on USE flags yet.  Although, I think such a thing would be possible with the right patches.

Issue #32, opened 2017-10-09
dev-qt/qtscript LTO build fix


pchome:
Based on https://bugs.webkit.org/show_bug.cgi?id=56088
```patch
--- a/src/3rdparty/javascriptcore/JavaScriptCore/jit/JITStubs.h 2017-10-09 14:40:55.050363361 +0300
+++ b/src/3rdparty/javascriptcore/JavaScriptCore/jit/JITStubs.h 2017-10-09 14:45:50.271354448 +0300
@@ -316,7 +316,7 @@
     EncodedJSValue JIT_STUB cti_op_to_primitive(STUB_ARGS_DECLARATION);
     EncodedJSValue JIT_STUB cti_op_typeof(STUB_ARGS_DECLARATION);
     EncodedJSValue JIT_STUB cti_op_urshift(STUB_ARGS_DECLARATION);
-    EncodedJSValue JIT_STUB cti_vm_throw(STUB_ARGS_DECLARATION);
+    EncodedJSValue JIT_STUB cti_vm_throw(STUB_ARGS_DECLARATION) __attribute__((used));
     EncodedJSValue JIT_STUB cti_to_object(STUB_ARGS_DECLARATION);
     JSObject* JIT_STUB cti_op_construct_JSConstruct(STUB_ARGS_DECLARATION);
     JSObject* JIT_STUB cti_op_new_array(STUB_ARGS_DECLARATION);

```

I don't tested it further but at least it builds w/ this patch in `/etc/portage/patches/dev-qt/qtscript/`

pchome:
Ok, I played a little with Qt-5.9.2 today and finally it full LTOed for me.

Here the installed packages list:
```sh
$ eix -I --only-names dev-qt/*
```
```sh
dev-qt/designer
dev-qt/linguist-tools
dev-qt/qdbus
dev-qt/qtchooser
dev-qt/qtconcurrent
dev-qt/qtcore
dev-qt/qtdbus
dev-qt/qtdeclarative
dev-qt/qtgraphicaleffects
dev-qt/qtgui
dev-qt/qtimageformats
dev-qt/qtlockedfile
dev-qt/qtmultimedia
dev-qt/qtnetwork
dev-qt/qtopengl
dev-qt/qtpaths
dev-qt/qtprintsupport
dev-qt/qtquickcontrols
dev-qt/qtquickcontrols2
dev-qt/qtscript
dev-qt/qtsingleapplication
dev-qt/qtsql
[...]

PR #48, opened 2017-10-19
Support more LTOed packages

pchome:
I have all packages built w/ this for now and no noticeable problems so far. The only `-flto` overrides left is (as have some build errors):
```sh
media-libs/alsa-lib
media-libs/flac
```

and this two (unstable w/ LTO):
```sh
net-misc/openssh
media-sound/jack2
```
Some other locally patched to be built w/ LTO or have additional flags.
So for now I have ~1K packages installed and still no any `-ffat-lto-objects` flags set.

### DIFF
--- a/sys-config/ltoize/files/make.conf.lto
+++ b/sys-config/ltoize/files/make.conf.lto
@@ -2,6 +2,14 @@
 #It's usually a good idea to set this to the number of hardware threads in your system
 FLTO="-flto=8"
 
+# https://bugs.gentoo.org/603594
+# Keep all static libraries unstripped.
+# This should allow you to omit almost all -ffat-lto-objects flags and build many more packages with LTO.
+# After enabling this you must rebuild all packages providing static libraries with LTO support.
+# Use $ `equery b -n $(find /lib* /usr/lib* -type f -name "*.a" 2>/dev/null)` | sort | uniq
+# or # `emerge -1va $(find /lib* /usr/lib* -type f -name "*.a" 2>/dev/null)` commands to get corresponding list.
+STRIP_MASK="*.a"
+
 #GRAPHITE contains Graphite specific optimizations and other optimizations that are disabled at O3
 #but don't influence the compiler's judgement.
 #These should be updated with GCC 7.2.1 (-ftree-loop-distribution may be enabled


InBetweenNames:
Oh wow, glad to see some upstream discussion about this -- thanks for linking the thread!  I understand now why `strip` causes the problem it sometimes does.

It looks like perhaps a Portage patch is in order to fix this.  According to the linked thread, `gcc-ranlib` needs to be run in order to fix the static libraries.  I'm going to check into that to see if I can attach a patch to the bug report.  It would be nice to have it handled properly.

InBetweenNames:
Okay, I was able to patch Portage to do it, but of course the patch will need work.

File /usr/lib64/portage/python3.6/ebuild-helpers/prepstrip (will vary from system to system) line 347:

~~~
if ! ${FEATURES_splitdebug} || ${RESTRICT_splitdebug} ; then
    ${STRIP} -g "${x}"
    ${CHOST}-ranlib "${x}" #add this here
fi
~~~

This did the trick for me.  Obviously there are a few things that need to be addressed:

* ${STRIP} seems to be defined but ${RANLIB} is not.  Not sure where STRIP gets defined by default, but it seems RANLIB should be too, or a way of getting it at least.
* Different Python versions and architectures

---

While that's being worked on, instead of having a blanket `nostrip`, what if we applied that only to the individual packages that are affected?

pchome:
`STRIP_MASK` is internal portage env_var AFAIK
and "*.a" mask applied only for static libs (preinstall `strip` don't applied to this files in other words)

Not-stripped `*.a` won't hurt your system match cause there is a few packages provided them and I believe  final size almost equal to `fat objects` . So why not to do things more complex and just use this `STRIP_MASK` variable until `objdump` (`strip` is alias(?)) will support plugins?

pchome:
> what if we applied that only to the individual packages that are affected?
Sure you can use `pkg/name STRIP_MASK="*.a"` or so for individual packages in your `packages.cflags` but you'll face build problems caused by packages not in the list and other packages will be false-positive reported as -fno-lto ones. I know what I'm talking about cause I've done this before: https://github.com/InBetweenNames/gentooLTO/issues/32#issuecomment-335237832 😉 .

pchome:
One more example to be clear:
`$ equery f binutils-libs | grep "\.a$"`
```
/usr/lib32/libiberty.a
/usr/lib64/libiberty.a
```
but you have `binutils-libs` marked as *knownworking* so every package using `-liberty` flag will fail to build (`sys-devel/prelink` in my case) and probably will be reported as `-fno-lto`.

But it builds fine for me w/ LTOed and nostripped `libiberty.a` and it takes time to understand why.
So I rejected idea to add something like `nostrip` into my config and decided to set `STRIP_MASK="*.a"` globally to avoid similar problems in future.

InBetweenNames:
Wow, thanks for the in-depth information @pchome !  I'm sufficiently convinced now that we should go ahead with your approach on this.  It really does seem like this workaround is required until `strip` has LTO support, just as you said.  I will merge this and update the `README.md` to reflect this information shortly.

PR #52, opened 2017-10-22
Use some overrides only for specified USE-flag

pchome:
There is possible to add some logic to `*FLAGS` ( https://github.com/vaeth/portage-bashrc-mv/issues/8#issuecomment-338497760 ) so we can check is USE-flag set etc. You probably may want to add similar logic for `BEGIN: genkernel no LTO needs for initramfs use of LVM + LUKS` block and other packages.


pchome:
Sorry, this feature needs more testing as `$(...command...)` triggered for all packages.

InBetweenNames:
Oh nice!  Just a small question: in `${IUSE//+}`, what is the `//+` intended to do?  Is it supposed to just substitute `+` if `IUSE` is unset or empty?

pchome:
`$ VAR="one +two zero"; echo "${VAR//+}"`
```sh
one two zero
```

InBetweenNames:
Thanks for the clarification -- LGTM

PR #53, opened 2017-10-24
qtscript:4 cannot be build with lto

soredake:


InBetweenNames:
Looks good to me, just a minor change: could you keep the original commented line and add the 5 slot to it?

soredake:
Fixed.

InBetweenNames:
Thanks for the contribution @soredake !

PR #54, opened 2017-10-24
dev-libs/wayland: LTO patch

pchome:


InBetweenNames:
Beautiful!

pchome:
Should be available upstream now. 1.15.0.

InBetweenNames:
Awesome.  We can get rid of the older userpatches now

PR #55, opened 2017-10-24
media-libs/x264: LTO patch

pchome:


InBetweenNames:
Nice!  Just curious, does the patch only work on the snapshot version or does it work on the latest one in the `/usr/portage` tree as well?

pchome:
(~)0.0.20170701(0/152) is the latest unstable snapshot in tree.

And due to http://git.videolan.org/?p=x264.git;a=blob;f=configure;h=f7b14d9d9730d9830e132e5d8ab50b7a3c066400;hb=HEAD#l961 this patch should work even for 9999 and maybe for prev versions too.

InBetweenNames:
I'll give it a try when I boot into Gentoo later today.  If it works on the latest testing x264, we may as well just apply your patch to that instead and that way there's less to maintain.  Nice work!  With all the user patches we've been getting, I may look into a better mechanism to merge them into `/etc/portage/patches`.

InBetweenNames:
Alrighty, finally got a chance to look at it.  So I see now why you included the ebuild itself: you needed the `epatch_user` functionality.  However, since we're supplying the ebuild ourselves, we could actually just add the patch to the `PATCHES` variable and forego `epatch_user` altogether.  What do you think about that change?  We could make it policy that if an ebuild is supplied, then the LTO patches must be applied in the ebuild itself.  Otherwise, it may be applied with `eapply_patch` (for out of tree ebuilds).

pchome:
Ok, you can change this PRs on your needs and I will follow policy in future ones.

InBetweenNames:
Sounds good!  Next time I'm doing my Gentoo stuff I will amend this as needed and merge it.

PR #56, opened 2017-10-24
(hardcore) dev-lang/spidermonkey:24: LTO patch

pchome:
The only package requires dev-lang/spidermonkey:0 in my system is  sys-auth/polkit.

So here is patched dev-lang/spidermonkey:24 for LTO support and patched sys-auth/polkit for dev-lang/spidermonkey:24 support.

Despite changes looks too major  I used this patches for a while w/o any noticeable issues.


InBetweenNames:
Lets test this thing out!

InBetweenNames:
I spoke too soon -- it looks like the LTO patches need some work.  In particular, the casts will need to be fixed up.  I've reverted it for now but I'm still interested in getting this merged.

PR #62, opened 2017-11-04
lto breaks rssguard

soredake:


InBetweenNames:
Hi soredake,

I was actually able to get this built without needing any overrides.  Just curious, did you implement the `STRIP_MASK` changes suggested recently?  @pchome showed conclusively that the `fat-lto-objects` necessity is often from a bug in `strip`.  The program here seems to run, too -- I just tested out the feed functionality and it works.

soredake:
```
cat /etc/portage/make.conf | grep STRIP_MASK
STRIP_MASK="*.a"
```
For me with `fat-lto-objects` it just shows nothing, then it breaks the db at exit and segfaults at next start.

InBetweenNames:
Confirmed, thanks!

PR #63, opened 2017-11-04
games-emulation/ppsspp: graphite breaks build

soredake:


InBetweenNames:
Excellent, thank you!

PR #70, opened 2017-11-21
Additional packages that have issues with LTO and/or Graphite....

preney:
This pull request contains additional packages that could not be emerged due to LTO and/or Graphite optimizations.


InBetweenNames:
Confirmed.  Thanks!

PR #71, opened 2017-11-21
sys-process/criu won't build without significant overrides

preney:
Package sys-process/criu requires overrides (including removal of -pipe for ld). The override list probably can be chopped down a bit --but this will allow the code to be installed.


InBetweenNames:
Confirmed, thanks!

PR #72, opened 2017-11-22
kde-apps/kreversi: compiles with LTO

pchome:


InBetweenNames:
Heh, I guess this one slipped by me.  Confirmed.  Thanks!

PR #73, opened 2017-11-28
Revert "(hardcore) dev-lang/spidermonkey:24: LTO patch"

InBetweenNames:
Will need to modify the patches to build properly


pchome:
FYI: recent `sys-auth/polkit` depends on newer `dev-lang/spidermonkey:52` which builds w/ LTO enabled.

USE flags:
`52.9.1_pre1(52)(17:03:48 30.07.18)(custom-cflags custom-optimization minimal system-icu -debug -test)`

PR #75, opened 2017-11-30
=media-libs/mesa-17.3.0_rc5: ICE with LTO

soredake:


InBetweenNames:
It looks like this is Nouveau specific, however I will merge this as we do not have a mechanism for making LTO workarounds conditional on USE flags at this time

InBetweenNames:
Thanks for the contribution!

pchome:
> It looks like this is Nouveau specific, however I will merge this as we do not have a mechanism for making LTO workarounds conditional on USE flags at this time

Are this tricks don't work :
`media-libs/mesa "has video_cards_nouveau ${IUSE//+} && use video_cards_nouveau && FlagSubAllFlags -flto*"`
?

same for `cpu_flags_x86_sse4_1` or so

InBetweenNames:
Oh, I like that!  I was thinking of something along these lines:

~~~
media-libs/mesa[nouveau] *FLAGS-=-flto*
*/ffmpeg[cpu_flags_x86_sse4_1] *FLAGS-=-flto*
~~~

It would require a few patches to make it work mind you.  I'll open an issue about this so we can get some more input.

PR #76, opened 2017-12-03
dev-lang/php: fix LTO for version >= 7.2

pchome:
*FLAGS+="-flto-partition=max" don't work for dev-lang/php:7.2 anymore


InBetweenNames:
It's interesting that global register variables conflict with LTO here.  Definitely worth some investigation in the PHP codebase.  Thanks!

Issue #78, opened 2017-12-06
Augment workarounds with USE-flag oriented exceptions


InBetweenNames:
Certain workarounds are only necessary in the presence of particular USE flags (see #75).
It would be nice to be able to specify workarounds that apply only in these cases, for example:

~~~
media-libs/mesa[nouveau] *FLAGS-=-flto*
*/ffmpeg[cpu_flags_x86_sse4_1] *FLAGS-=-flto*
~~~

This would require some patches to make it work.  One approach that works right away without any patching at all is:

~~~
media-libs/mesa "has video_cards_nouveau ${IUSE//+} && use video_cards_nouveau && FlagSubAllFlags -flto*"
~~~

(courtesy of @pchome)

As contributors of this repo, I'd like to ask you what the best approach for doing this would be.  I'll leave this thread open for any discussion or ideas on this.

pchome:
The simpliest way is to add */etc/portage/bashrc.d/0000-lto-overlay-helpers.sh*:
```bash
#!/bin/bash

# Helpers
# for LTO overlay [https://github.com/InBetweenNames/gentooLTO]
# and package.cflags [https://github.com/vaeth/portage-bashrc-mv]

# based on eutils.eclass:use_if_iuse()
lto_use_if_iuse() {
        [[ ${#} -eq 1 ]] || die "Invalid args to ${FUNCNAME}()"
        local flag=${1}
        local liuse=( ${IUSE} )

        has "${flag}" "${liuse[@]#[+-]}" || return 1
        use $1
}

no_lto_if_use() {
        lto_use_if_iuse $1 && FlagSubAllFlags -flto*
}

no_graphite_if_use() {
        lto_use_if_iuse $1 && FlagSubAllFlags ${GRAPHITE}
}

drop_flag_if_use() {
        [[ ${#} -eq 2 ]] || die "Invalid args to ${FUNCNAME}()"
        lto_use_if_iuse $1 && FlagSubAllFlags $2
}

...
```
and then `media-libs/mesa "no_lto_if_use video_cards_nouveau"`
or `media-libs/mesa "drop_flag_if_use video_cards_nouveau -flto*"`

pchome:
Also you can simplify all this common `*FLAGS-=-SGALF*` with
```bash
...
noLTO() {
        FlagSubAllFlags -flto*
}

noGRAPHITE() {
        FlagSubAllFlags ${GRAPHITE}
}

fatLTO() {
[...]

PR #83, opened 2017-12-09
media-libs/mesa: disable lto only with video_cards_nouveau

soredake:
https://github.com/InBetweenNames/gentooLTO/issues/78


InBetweenNames:
Thanks!

PR #86, opened 2017-12-17
Return keywords to ltoize ebuild and lower gcc-config version require…

soredake:
@InBetweenNames


InBetweenNames:
Ahh, good catch!  I didn't see 1.9.1 there last night.  Wonderful!

PR #99, opened 2018-02-15
app-misc/vlock: disable LTO

Francesco149:
Title: app-misc/vlock: fails with ```gcc: error trying to exec '/usr/libexec/gcc/x86_64-pc-linux-gnu/7.3.0/collect2': execv: Argument list too long``` unless I disable LTO completely.

from the error log it seems like it's more of a build system problem than LTO itself

full log https://gist.github.com/anonymous/cd87a7af30314bbb9d90ec49073294e4


InBetweenNames:
Yup I can reproduce this, and thank you for attaching your build log.  LGTM!

PR #118, opened 2018-05-22
app-emulation/wine*: can be built with graphite

pchome:
`app-emulation/wine-staging` compiles for me with GCC 8.1.0 and graphite enabled.


InBetweenNames:
Thanks!

PR #121, opened 2018-05-29
dev-db/mysql-connector-c: Added LTO exception to allow compilation

mylanconnolly:
Noticed a failure for dev-db/mysql-connector-c when utilizing LTO. Added exception and compilation proceeds normally. Please let me know if there are any issues.

Thanks!


InBetweenNames:
Thanks!

PR #134, opened 2018-07-19
sys-libs/libse{pol,linux,manage}, app-crypt/efitools, gnome-base/gnom…

### LINKED ISSUE #131: A couple more packages that don't seem to work with LTO
sys-libs/libse{pol,linux,manage}, app-crypt/efitools, gnome-base/gnome-keyring, dev-libs/weston, app-crypt/veracrypt, dev-qt/qtscript
I currently have `*FLAGS-=-flto*` set for all of them except for efitools, which has `*FLAGS+=-ffat-lto-objects`, which makes them compile and, seemingly, work. I am not sure if there's a better way to handle these packages though.

monoidic:
…e-keyring, dev-libs/weston, app-crypt/veracrypt

fix #131

dev-qt/qtscript seems to be in already.


InBetweenNames:
Thanks for the contribution!

PR #153, opened 2018-10-11
net-misc/lksctp-tools: disable LTO

zeule:


InBetweenNames:
Tested and confirmed.  Thanks for the contribution!

PR #160, opened 2018-10-29
sys-apps/exa: fails to link to git2 with fat LTO

Althorion:
`sys-apps/exa` fails to build with fat LTO, for the same reasons I guess that `cargo` does.


InBetweenNames:
Thanks!

PR #174, opened 2018-11-06
sys-devel/llvm with SEMINTERPOS makes DXVK crash

faultyme:
Compiling llvm with semantic flag causes wine with DXVK either freeze or crash. Steam play/Proton is also affected, however native vulkan games on Linux is unaffected like RoTTR.
My gpu: AMD VEGA 64


InBetweenNames:
Thanks for the contribution!

Issue #175, opened 2018-11-07
cannot emerge polybar with CMAKE_MAKEFILE_GENERATOR=ninja


aw1cks:
With ninja: [link](https://pastebin.com/raw/cpt7J0Wg)
With emake: no errors, emerges without problem.
I just created /etc/portage/package.cflags/polybar.conf with "x11-misc/polybar CMAKE_MAKEFILE_GENERATOR=emake". Maybe it's a good idea to add a note like with kinfocenter?

aw1cks:
It's done. See the PR [here](https://github.com/InBetweenNames/gentooLTO/pull/176) :)

aw1cks:
Just an afterthought: given that I have created a new section in ltoworkarounds.conf in my PR, perhaps it would be prudent to remove the note about kinfocenter and simply add it to the file straight away.

InBetweenNames:
Thanks!  I think you did the right thing -- we can move these helpful workarounds into a separate file if we get more than, say, five of them.

PR #176, opened 2018-11-07
x11-misc/polybar: use emake instead of ninja

aw1cks:
Polybar will not build with ninja; added line to use emake. See [issue](https://github.com/InBetweenNames/gentooLTO/issues/175)


InBetweenNames:
Thanks for the contribution!

Issue #179, opened 2018-11-10
thermald failing, needs *FLAGS-=-flto*


0rmi:
With LTO disabled it compiles fine, with it, failure. -ffat-lto-objects doesn't help.
emerge --info: https://pastebin.com/x3bLtyfq 
build.log: https://pastebin.com/TKEwqSBx

aw1cks:
After reproducing the issue on my laptop, I created a [PR](https://github.com/InBetweenNames/gentooLTO/pull/180).

InBetweenNames:
Actually, I managed to get it to build with LTO with this:
~~~
sys-power/thermald *FLAGS+=-Wl,--no-fatal-warnings *FLAGS+=-Wno-error
~~~

Could you try this out?

0rmi:
Works for me. Thanks!

InBetweenNames:
Excellent.  Will use this workaround instead and close the issue.

PR #180, opened 2018-11-10
sys-power/thermald - disabled LTO to allow build

aw1cks:
See [issue](https://github.com/InBetweenNames/gentooLTO/issues/179)


InBetweenNames:
Thanks for the PR -- in the issue we decided to go with another approach rather than disabling LTO, and the fix should be committed.

PR #191, opened 2018-11-16
app-doc/doxygen: added to the ninja blacklist.

### LINKED ISSUE #190: app-doc/doxygen (1.8.14-r1) fails to build with CMAKE_MAKEFILE_GENERATOR=ninja
ninja fails because of `ninja: error: build.ninja:1915: multiple rules generate generated_src/configvalues.h [-w dupbuild=err]`

the workaround of setting `NINJAOPTS="-w dupbuild=warn"` through `ninja-utils.eclass` fixes the make phase, but not the install.

further research shows that `cmake-utils.eclass` does not include NINJAOPTS at the install phase.

so, for now `CMAKE_MAKEFILE_GENERATOR=emake` is necessary to build.

The-Bishop:
Fixes #190


InBetweenNames:
Thanks for the contribution!

Issue #193, opened 2018-11-16
dev-scheme/guile (2.2.3) fails to build with LTO


The-Bishop:
`*FLAGS+=-ffat-lto-objects` was not enough, so had to use the big knife to cut out LTO alltogether to make the package build.

PR adds `*FLAGS-=-flto*` to `sys-config/ltoize/files/package.cflags/ltoworkarounds.conf`

InBetweenNames:
Thanks!

PR #194, opened 2018-11-16
dev-scheme/guile: disable LTO for this package

The-Bishop:


The-Bishop:
Fixes #193

InBetweenNames:
Confirmed and merged, thanks!

Issue #189, opened 2018-11-15
dev-libs/libsigsegv-2.12-r1 requires *FLAGS-=-flto*


jelinekto:
[build.log](https://github.com/InBetweenNames/gentooLTO/files/2585008/build.log)
[emerge-info.txt](https://github.com/InBetweenNames/gentooLTO/files/2585009/emerge-info.txt)

InBetweenNames:
Confirmed -- could you make a PR out of this?

pchome:
Looks like `dev-libs/libsigsegv` trying to mix
```
sys-kernel/linux-headers-4.19 (/usr/include/asm/sigcontext.h)
sys-libs/glibc-2.27-r6 (/usr/include/bits/sigcontext.h)
```
Probably configuration script fault, and can be fixed by removing `-flto*` for either configuration script or particular test (see/compare `config.log`).

PR #195, opened 2018-11-16
dev-libs/libsigsegv: disable LTO

jelinekto:
See #189


InBetweenNames:
Thanks for the contribution!

Issue #140, opened 2018-08-02
app-shells/bash & sys-libs/readline false positive with LTO & O3 on configure MUST_REINSTALL_SIGHANDLERS


zhuyifei1999:
Reproduced on latest ~amd64:
=sys-devel/gcc-7.3.0-r3 USE="cxx fortran multilib nls nptl openmp pch pgo pie sanitize ssp vtv"
=app-shells/bash-4.4_p23
=sys-libs/readline-7.0_p5

`CFLAGS='-march=native -O3 -flto=4 -fuse-linker-plugin -pipe -ggdb -Wl,--as-needed -Wl,--hash-style=gnu'`
`LDFLAGS='-Wl,--as-needed -Wl,--hash-style=gnu'`
(Copied from `make.conf.lto`, with LTO enabled (4 threads) and Graphite disabled, and add `-ggdb`)

The symptom: in a maximized terminal window with bash running,
* force a new prompt (by pressing enter on a blank prompt)
* type a character
* un-maximize it (as long as the horizontal size is decreased)
* type another character (can be the same)
* maximize again

In this state, long bash commands (typed or scrolled in bash history) will, somehow, not wrap to new lines, but overwrite the current line. 

By `strace` debugging, it is visible that when bash echoes the commands, it outputs carriage returns `\r` where it expect the line to wrap in the smaller terminal window; however, because that incorrect expectation of line length, it starts 'wrapping' by `\r` before the line is exhausted, and `\r` in this case incorrectly wraps to the beginning of the current line.
It is also seen in `strace` that on each terminal resize `SIGWINCH` is correctly delivered to the bash process.

By `gdb` debugging, using an [a gdb script](http://palves.net/list-active-signal-handlers-with-gdb/) to check the active signal handler for `SIGWINCH` (28), it is seen that on a new prompt the handler is `rl_sigwinch_handler in section .text of /lib64/libreadline.so.7`, but after the signal is delivered it is changed to `sigwinch_sighandler in section .text of /bin/bash`.

[...]

PR #196, opened 2018-11-17
sys-libs/readline: patch is now in upstream. good job improving the ecosystem!

wolfwood:
#140


InBetweenNames:
Accepted as of #226!

Issue #197, opened 2018-11-18
media-sound/mpc: building with -fipa-pta causes hang on exit


jelinekto:
This happens on both 0.30 and 0.31.
[emerge-info.txt](https://github.com/InBetweenNames/gentooLTO/files/2592910/emerge-info.txt)

InBetweenNames:
Could you make a PR for this?

PR #202, opened 2018-11-21
media-sound/mpc: building with -fipa-pta causes hang on exit

jelinekto:
See #197


InBetweenNames:
Thanks for the contribution!

Issue #209, opened 2018-12-15
sys-apps/acl needs LTO to be desactivated


Aphypnise:
Well, I think I spent two or three days, trying to fix audio issues and acl permissions, with all this consolekit / elongind / polkit undocumented mud.

I found the issue on the gentoo tracker, ACL attribution will always fail if compiled with LTO, at least for the last version in the main gentoo tree.

I will perhaps run some tests, to be sure that's really and completely LTO that doesn't work.

Cf : https://bugs.gentoo.org/667372

InBetweenNames:
I actually remember having some trouble with this package a long time ago -- based on the bug you've posted that's confirmation enough for me there's a problem.  Would you be willing to create a pull request for this?

Aphypnise:
Yeah no problem, should be done tomorrow.

InBetweenNames:
Merged, thanks for your contribution!

PR #210, opened 2018-12-17
sys-apps/acl fails to set ACLs if built with LTO.

Aphypnise:
See #209 

Basically, it builds fine but we cannot set any ACLs at all.


InBetweenNames:
Thanks!

PR #212, opened 2018-12-24
dev-libs/glib: -fipa-pta causes very strange behavior (affecting dbus, for example)

asaparov:
Added glib workaround for -fipa-pta. This is not a compilation error. But for some reason, -fipa-pta changes behavior of dbus (which depends on glib) such that consolekit, NetworkManager, KDE, etc can no longer correctly work with dbus, affecting the entire system.


InBetweenNames:
This is very strange!  On my system it seems to be working OK -- however I'm okay with accepting this PR so long as there's a small one line comment explaining the need for the workaround.  Occasionally I review the exclusion list and throw out obsolete entries.  Does that sound okay to you?

asaparov:
Oh that's very odd. Hmm, maybe we shouldn't add it for everyone yet? Maybe it has something to do with the fact that I'm running a mostly Ofast system? In any case, I added the comment.

donbcd:
I can somehow confirm this, but there must be more to it. A few weeks ago (before I used lto-overlay myself) I played around with -fipa-pta, -fno-semantic-interposition, and -falign-functions and ran into exactly the same problem with glib and dbus. Because I ran out of time and it was just a try anyways I rolled back to my working configuration and forgot about it.
But now, with the lto-overlay, everything works fine after rebuilding world. Yet, this time I did not add the -falign-functions, may that is the missing part? Also, as dbus not working is a very severe problem, I think more people would have complained if it were a general issue.

asaparov:
Interesting. I just tried rebuilding `glib` without -falign-functions but -fipa-ita still causes the weird behavior. Maybe other flags are causing the issue? Or maybe I need to rebuild more packages with falign-functions? Were you using other flags, like O3, Ofast, ffast-math a few weeks ago?

donbcd:
-ftree-vectorize -fipa-pta -fno-semantic-interposition -march=skylake -O3 -falign-functions=32
I did not have lto enabled for dbus or glib
I did not rebuild world back then, so it can really be the problem that some other packages need to be rebuilt in order to make glib and dbus work again.

InBetweenNames:
@asaprov is this issue still affecting you?

asaparov:
Yes, I just tried building glib with -fipa-pta again, and consolekit still can't connect to dbus. Was there a recent change/update that could have fixed this?

But I'm still not sure why I have this issue but others don't.

gcs-github:
GCC 8.3.0 was just released, with some fixes in its InterProcedural Analysis code (the IPA in -fipa-pta). Might be worth re-testing after compiling with it if you can. Additionally, binutils 2.32 made it to ~arch, so you might want to upgrade there as well.

asaparov:
Thanks for the heads up. I just tested with gcc 8.3.0 and binutils 2.32 and rebuilt glib with -fipa-pta and I ran into the same problem (console-kit-daemon doesn't correctly connect to dbus). =/

InBetweenNames:
#257 -- anyone affected by this issue should read that issue too.  I've disabled `-fipa-pta` globally until this gets resolved.  Anyone manually using `-fipa-pta` (i.e., not sourcing `make.conf.lto`) should update their configs manually.

InBetweenNames:
OK -- with GCC 9.1.0 out and the `-fipa-pta` regressions fixed, I'm able to finally close this.  If it needs to be reopened, just comment and I'll do so.

PR #214, opened 2018-12-28
dev-scheme/racket: fails to build with LTO and -fipa-pta

aphypnise:
Signed-off-by: Andrew Attali <me@aphypnise.eu>

Racket needs to be built without LTO and -fipa-pta, since some references are used by the racket compiler during installation. This prevents segfaults and undefined references.


InBetweenNames:
Thanks for the contribution!

Issue #204, opened 2018-11-28
dev-java/openjdk fails to build with LTO


mateuszmandera:
With *FLAGS-=-flto* it builds fine.

[build.log](https://github.com/InBetweenNames/gentooLTO/files/2625821/build.log)
[emerge-info.log](https://github.com/InBetweenNames/gentooLTO/files/2625828/emerge-info.log)

InBetweenNames:
I have actually been having trouble with the openjdk packages recently.  Unfortunately, I'm preparing for my thesis proposal on December 3rd, so I won't be able to investigate any issues until then.  I can review pull requests at least, if they are simple.

aphypnise:
Can confirm, tried with -ffat-lto-objects but it's uniquely without LTO that the compilation works. I'm going to send a PR.

InBetweenNames:
Reproduced and merged the PR.  Thanks!

PR #215, opened 2018-12-29
dev-java/openjdk: fails to build without LTO off.

aphypnise:
See #204

Signed-off-by: Andrew Attali <me@aphypnise.eu>


InBetweenNames:
Thanks for the contribution!

PR #218, opened 2019-01-06
Fix GCC version number in warning message

SpontaneousOverthrow:
I am not sure, but GCC version in warning message is not equal to "${ACTIVE_GCC}" -lt 8.2.0


InBetweenNames:
Thanks for the contribution!

PR #219, opened 2019-01-07
net-fs/autofs: no-semantic-interposition segfaults

nonno-cicala:
With -fno-semantic-interposition builds fine but crashes with:

> automount[9179]: segfault at 0 ip 00007ff50bd65488 sp 00007fffa36d3070 error 4 in lookup_file.so[7ff50bd4c000+36000]


InBetweenNames:
Confirmed, thanks!

PR #227, opened 2019-01-15
Added workaround for sys-apps/fwupd

### LINKED ISSUE #225: sys-apps/fwupd Does not work with LTO
I had to build `fwupd` with the following options (no LTO) for it to work properly
```
CFLAGS="-march=native -O3 ${SEMINTERPOS} ${GRAPHITE} ${IPA} -pipe -falign-functions=32"
CXXFLAGS="${CFLAGS}"
LDFLAGS="${LDFLAGS} -Wl,--hash-style=gnu"
```
Otherwise it fails it start and all it's commands fail to execute properly.

bemeurer:
Fixes #225


InBetweenNames:
Thanks for the contribution! :)

PR #230, opened 2019-01-16
Added workaround for dev-python/grpcio

### LINKED ISSUE #229: dev-python/grpcio Fails to build with LTO
Package fails during linking, disabling LTO fixes it.

bemeurer:
Fixes #229


InBetweenNames:
Thanks for testing -- will use the patch in the issue #229

Issue #129, opened 2018-06-27
lto1: internal compiler error when building net-p2p/retroshare-0.6.4 


zeule:
lto1: internal compiler error: in get_odr_type, at ipa-devirt.c:2098

Probably an exception is needed.

InBetweenNames:
Actually, I can't even get this to compile with GCC 8.1.0:

~~~
grouter/p3grouter.cc:543:65:   required from here
/usr/lib/gcc/x86_64-pc-linux-gnu/8.1.0/include/g++-v8/bits/stl_tree.h:452:21: error: static assertion failed: comparison object must be invocable with two arguments of key type
       static_assert(__is_invocable<_Compare&, const _Key&, const _Key&>{},
~~~

InBetweenNames:
I'm not sure which compiler you're using, but have you tried doing no-lto or fat-lto with it?

InBetweenNames:
Oh great! Would you mind creating a PR to add `net-p2p/retroshare` as no-lto?

zeule:
Please (#130).

InBetweenNames:
Thanks for the contribution! :)

PR #236, opened 2019-01-24
Added workaround for sci-visualization/paraview

corrmaan:
Similar to https://github.com/InBetweenNames/gentooLTO/issues/129, compilation fails with error:
`lto1: internal compiler error: in get_odr_type, at ipa-devirt.c:2098`
Removing LTO allowed it to compile.


corrmaan:
Just tried `sci-visualization/paraview *FLAGS-=-fipa-pta` and compilation failed with the same error.

gcs-github:
Looking into it, it seems that much the Inter Procedural Analysis code is enabled through -flto, and that this part of GCC still has a few ICEs lying around.

A possibly related GCC patch (I didn't analyse it thoroughly enough to determine if it would fix this particular issue): http://gcc.1065356.n8.nabble.com/Fix-another-ICE-in-ipa-devirt-td1535860.html

InBetweenNames:
Confirmed and accepted -- thanks for the contribution!  @gcs-github thanks for the interesting link

PR #242, opened 2019-02-05
Added -fipa-pta workaround for libwebp.

asaparov:
media-libs/libwebp: -fipa-pta causes browsers (tested with firefox and chrome) to incorrectly display webp images.


InBetweenNames:
Confirmed, thanks!

InBetweenNames:
Problem confirmed in `dev-qt/qtgui` too, thanks to your PR.

PR #246, opened 2019-02-14
media-gfx/shotwell: Disable LTO

bemeurer:
Shotwell does not build with LTO enabled.


InBetweenNames:
Thanks for the contribution!

PR #251, opened 2019-02-24
dev-qt/designer: Disable LTO

PeeJay:
kde-frameworks/kjsembed will fail to build otherwise. The QtUiTools:: class is MIA during linking.

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -12,6 +12,7 @@ sys-apps/exa *FLAGS+=-ffat-lto-objects # fails to link against git2 functions
 sys-apps/bat *FLAGS+=-ffat-lto-objects # fails to link against git2 functions
 
 #Packages which cannot be built with LTO at all (previously was nolto.conf)
+dev-qt/designer *FLAGS-=-flto*
 dev-util/perf *FLAGS-=-flto*
 media-libs/mesa *FLAGS-=-flto* #*FLAGS-=-Wl,--as-needed # strange LTO linking bug that causes pthreads to be thrown out early under LTO (#50)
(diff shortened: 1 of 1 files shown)


pchome:
`$ cat /var/db/pkg/dev-qt/designer-5.11.3/CFLAGS`
```
-march=native -O3 -pipe -fgraphite-identity -ftree-loop-distribution -floop-nest-optimize -fipa-pta -flto=3 -fuse-linker-plugin -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu
```
`cat /var/db/pkg/kde-frameworks/kjsembed-5.55.0/CFLAGS`
```
-march=native -O3 -pipe -fgraphite-identity -ftree-loop-distribution -floop-nest-optimize -fipa-pta -flto=3 -fuse-linker-plugin -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu
```

InBetweenNames:
I actually have designer emerged without workarounds needed.  Could you post a complete `build.log`?

PeeJay:
See bug report #252

zeule:
Have kde-frameworks/kjsembed-9999 emerged without problems.

funghetto:
I had the same problem because I wasn't using >=sys-apps/portage-2.3.52 that had the patch to _restore the index of any static archive that has been stripped._

PeeJay:
> I had the same problem because I wasn't using >=sys-apps/portage-2.3.52 that has the patch to _restore the index of any static archive that has been stripped._

That was it, thanks.
Should a dependency on >=sys-apps/portage-2.3.52 be added to sys-config/ltoize so others don't have the same trouble?

InBetweenNames:
@PeeJay that seems like a good idea.  I'll update it accordingly when I'm able to.  (will leave this PR open until then)

InBetweenNames:
Fixed in `master`.  `sys-config/ltoize` now depends on a recent Portage (or portage-mgorny).

PR #256, opened 2019-03-09
x11-wm/bspwm: IPA workaround

insanitywholesale:
the latest version of bspwm wouldn't compile on my system without this workaround


InBetweenNames:
Confirmed, thank you!

Issue #264, opened 2019-03-28
dev-libs/isl-0.21 builds with -flto but fails with -flto=${NTHREADS}


jelinekto:
I'm guessing this is the important bit.
`cc: error trying to exec '/usr/libexec/gcc/x86_64-pc-linux-gnu/8.3.0/collect2': execv: Argument list too long`

[emerge-info.txt](https://github.com/InBetweenNames/gentooLTO/files/3015701/emerge-info.txt)
[build.log](https://github.com/InBetweenNames/gentooLTO/files/3015711/build.log)

gcs-github:
This looks like a limits problem. Do you have any non-default settings in /etc/security/limits.conf by any chance? Or something that would normally impact the default output or `ulimit -s`?
Can you CTRL+Z to stop emerge in the middle of the compilation, then run `ps aux` to find one of the compilation process IDs (make or gcc for example), and tell us what the output of `/proc/<pid_of_that_process>/limits` is?

ellisonpatterson:
Same issue.
ulimit -s
`8192`
limits.conf is unmodified.

gcs-github:
I'm reproducing. isl-0.20 compiled fine. 0.21 fails with the error mentioned in this issue. The diff between the two versions is 80k lines, so I'm not yet sure what's causing this exactly. I'm now suspecting a change in the package's build system.

gcs-github:
A `package.cflags` workaround:

```
dev-libs/isl /-flto=*/-flto=1
```

This will only slow down the compilation. The performance of the isl lib shouldn't be impacted.

gcs-github:
Turns out gentoo upstream has now masked isl-0.21 for this exact problem. For anyone using the workaround above, I think it's safe to add `=dev-libs/isl-0.21` to your local package.unmask

See: https://bugs.gentoo.org/682018

InBetweenNames:
Confirmed -- thanks @gcs-github for the report and for the workaround.  The PR has been merged.

ionenwks:
I haven't tried yet but seems this particular problem has been sorted by gentoo with a gcc patch.

https://bugs.gentoo.org/682018

gcs-github:
Yep. That was fast :)

gcs-github:
For some reason, the package is still masked in upstream Gentoo, but I can confirm that the problem is gone on my end.

ionenwks:
It got unmasked just now, just needed some time.

[...]

PR #265, opened 2019-03-28
dev-libs/isl: work around execv argument list too long upon linking

gcs-github:
Workaround for #264


InBetweenNames:

PR #274, opened 2019-04-05
add workaround for media-libs/faad2

huglovefan:
this fixes off-sounding aac playback in deadbeef (from deadbeef-overlay) when faad2 is compiled with -O3


InBetweenNames:
Confirmed, thanks for the contribution!

PR #272, opened 2019-04-03
media-libs/flac: disables -O3 default -fipa-cp-clone as it breaks flac -d with LTO

gcs-github:
Closes #270

PR #277, opened 2019-04-06
media-libs/flac: re-introduce #272

gcs-github:
Revert the accidental reversal of #272 which occurred in 3008105fff17b9bdf2efa34cb741c12c622b5bd9

Reported by @huglovefan


InBetweenNames:
My bad!  I should have been more careful before committing that change.  Thanks!

Issue #275, opened 2019-04-05
gnome-base/nautilus and its libs don't like LTO


ElDavoo:
Hi all,
Sorry if this issue is too generic, but gnome-base/nautilus-3.30.5 does not "start" with LTO.
See this forum post, not made by me but with exactly my issue, which explains everything in more detail. https://forums.gentoo.org/viewtopic-p-8324434.html .
I tried recompiling it without gentooLTO's optimization but nothing changes, so the issue should be in one of his dependencies.

ElDavoo:
the "bad" commit is https://gitlab.gnome.org/GNOME/nautilus/commit/a9c446afe262b7989bc17ef1f9ab19931db8dede

ElDavoo:
Ok, so the solution is to remove -fno-semantic-interposition from app-misc/tracker.

corrmaan:
Thanks for digging into this ElDavoo, using
`app-misc/tracker *FLAGS-=-fno-semantic-interposition`
in my LTO workarounds, then rebuilding tracker then nautilus, did the trick. I didn't try undoing that commit you mentioned.
I propose that you create a PR for this in `ltoworkarounds.conf` referencing that thread I opened ;)

InBetweenNames:
Confirmed and merged.  Thank you!

PR #279, opened 2019-04-08
app-misc/tracker: remove seminterpos

ElDavoo:
app-misc/tracker makes gnome-base/nautilus deadlock on start if it is compiled with ${SEMINTERPOS}.
Take a look at https://forums.gentoo.org/viewtopic-p-8324442.html and https://github.com/InBetweenNames/gentooLTO/issues/275 for more details.


InBetweenNames:
Excellent work, thanks!

PR #287, opened 2019-05-04
app-portage/lto-rebuild: Fix bashisms in /bin/sh script

automorphism88:
The script which this ebuild installs has a `/bin/sh` shebang, but uses two "bashisms" which cause problems if `/bin/sh` is dash. This can be verified either by running `dev-util/checkbashisms` on the script or by attempting to use the script with dash, where the use of `==` for comparison will cause errors and `echo -e` will not behave as expected.

An alternative fix would be to change the shebang to `/bin/bash`, but this is just the sort of simple script that doesn't need any bashisms.

Note that I only replaced `echo` with `printf` where the `-e` option was used, in order to make the changes as minimal as possible while still fixing the issue, but `printf` should generally be preferred over `echo` in shell scripts due to its greater portability.

### DIFF
--- a/app-portage/lto-rebuild/files/lto-rebuild
+++ b/app-portage/lto-rebuild/files/lto-rebuild
@@ -14,9 +14,9 @@ if [ "$#" -eq 0 ] || [ "$1" = "-h" ]; then
 	echo "command will fail.  You can use -l to manually intervene."
 	echo
 	echo "Options:"
-	echo -e "-h\tDisplay this help"
-	echo -e "-l\tOnly list the packages that would be rebuilt and their slots"
-	echo -e "-r\tRebuild the packages using emerge (will ask for confirmation)"
+	printf -- "-h\tDisplay this help\n"
+	printf -- "-l\tOnly list the packages that would be rebuilt and their slots\n"
+	printf -- "-r\tRebuild the packages using emerge (will ask for confirmation)\n"
 
 else
 	if [ "$1" = "-l" ]; then
--- a/app-portage/lto-rebuild/files/lto-rebuild
+++ b/app-portage/lto-rebuild/files/lto-rebuild
@@ -1,6 +1,6 @@
 #! /bin/sh
 
-if [ "$#" -eq 0 ] || [ "$1" == "-h" ]; then
+if [ "$#" -eq 0 ] || [ "$1" = "-h" ]; then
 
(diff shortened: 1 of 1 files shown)


app-portage/lto-rebuild/files/lto-rebuild
```diff
@@ -14,14 +14,14 @@ if [ "$#" -eq 0 ] || [ "$1" == "-h" ]; then
 	echo "command will fail.  You can use -l to manually intervene."
 	echo
 	echo "Options:"
-	echo -e "-h\tDisplay this help"
-	echo -e "-l\tOnly list the packages that would be rebuilt and their slots"
-	echo -e "-r\tRebuild the packages using emerge (will ask for confirmation)"
+	printf "-h\tDisplay this help\n"
```
huglovefan: you need to use `printf --` if the format string starts with a dash

```
% printf "-h\tDisplay this help\n"
dash: 1: printf: Illegal option -h
% printf -- "-h\tDisplay this help\n"
-h      Display this help
```

automorphism88: Good catch, fixed

InBetweenNames:
Thank you!  And thank you for the heads up on `dev-util/checkbashisms`, I will make sure to use that in the future before pushing.  Goes to show how much we rely on Bash these days eh?

InBetweenNames:
FYI: A test has been added to ensure this does not happen again!

automorphism88:
> Thank you! And thank you for the heads up on `dev-util/checkbashisms`, I will make sure to use that in the future before pushing. Goes to show how much we rely on Bash these days eh?

`dev-util/shellcheck` is really useful for checking shell scripts more generally. It looks at the shebang so it can use different rules for a bash versus sh script. But it does a lot more than just check for bashisms. And when it produces a false positive (e.g., you really did mean to single quote what looks like a variable expansion that should have been double quoted) you can add a directive that the shell will interpret as a comment but will prevent shellcheck from flagging it again. It requires a bunch of Haskell build-time dependencies but there's a `-bin` version to avoid that if you're so inclined.

And yeah, actually, Gentoo in particular is the only distro where using dash as `/bin/sh` has caused me problems, mainly because of poorly written upstream configure scripts. But the issues are generally easy to fix with a one-line patch when they do occur.

PR #318, opened 2019-05-15
app-portage/lto-rebuild: rewrote the script to make it more clean and readable

beresk137:
Got rid of loop while searching for .a files. Now parsing output of readelf instead of getting the filename from find — but we are relying on readelf output format anyway.
Also edited older ebuilds for sys-config/ltoize: nothing new in terms of business logic, just improved readability a bit.


InBetweenNames:
Wow this is great!!  Thank you!

PR #320, opened 2019-05-15
make.conf.lto

sandikata:
i would like to have removed source make.conf.lto.defines from make.conf.lto, as it cause issues with paludis package manager, and anyway makes managing simpler by sourcing both confs in make.conf directly.

Title: <category>/<package>: <description>

Ensure that any added workarounds have a comment explaining the compilation error which requires it.
Any removed workarounds should be moved to the fixed section.
If an issue exists, please link it.
See CONTRIBUTING.md for more information.

### DIFF
--- a/make.conf.lto
+++ b/make.conf.lto
@@ -1,0 +1,92 @@
+#This file contains the optimization defaults that GentooLTO uses
+#This includes O3, Graphite, LTO, and -fipa-pta
+#More will be added in the future.
+#Sourcing this file will enable you to transparently receive these new settings
+#without intervention.  If this is not desirable, see the file make.conf.lto.defines for custom configurations.
+
+#To use this file, first define NTHREADS to the number of threads you want the LTO
+#processes to use on your system at the top of your make.conf:
+
+#NTHREADS="12" 
+
+#A good number is the number of hardware threads you have available on your system.
+#After this, but before any other variables are defined, source this file directly:
+
+#source make.conf.lto
+
+#Then, afterwards, define your CFLAGS with your own options:
+
+#CFLAGS="-march=native ${CFLAGS} -pipe"
+
+#This enables your CFLAGS to inherit the default flags that GentooLTO uses.
+#Note that you may want to enable -falign-functions=32 if you use an Intel processor (Sandy Bridge or later).
+#See issue #164 for details.
+
+#Next, set your CXXFLAGS to CFLAGS:
+
+#CXXFLAGS="${CFLAGS}"
+
+#Your CXXFLAGS should be a superset of your CFLAGS.  Most people will never need to add anything extra to their CXXFLAGS.
+
+#Your LDFLAGS should contain *ALL* of your CFLAGS for LTO to work.  It must receieve all optimization options, march, etc.
+#package.cflags takes care of this for you--no need to manually add your {C,CXX}FLAGS here.
+#It's usually a good idea to enable -Wl,--hash-style=gnu as well to help find packages which don't respect LDFLAGS
+#Ensure that your profile's LDFLAGS are respected by including them first.
+
+#LDFLAGS="${LDFLAGS} -Wl,--hash-style=gnu"
+
+#NOTE: your profile likely contains -Wl,--as-needed and -Wl,-O1.
+#The -Wl,-01 here is a separate option sent to the linker that optimizes shared object binary sizes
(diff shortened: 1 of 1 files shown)


InBetweenNames:
It would be nice to support Paludis.  I have a refactor planned for GentooLTO which I think could take these needs into account.  For now though, I don't want to break any existing setups that source `make.conf.lto` directly.

javashin:
I used Paludis In The Past but is not being updated anymore and is years behind portage eapi 6,7 not supported , why care about paludis if it will break on every boost ruby python new version ? i dont see anyone using paludis ..... @sandikata  you are really brave my friend , one thing is live in the edge in software means but lives in the past is worst , maybe gentoo is not for you anymore try exherbo they support paludis until that distro is dead.

sandikata:
I don't care about your opinnion mate! 
Paludis have a lot of advantages compared to portage, main advantage is speed and multithreading, which is important think for me.

javashin:
LOL

javashin:
> I don't care about your opinnion mate!
> Paludis have a lot of advantages compared to portage, main advantage is speed and multithreading, which is important think for me.

Ok Ok ... Calm Down . Im Sorry If I Bother You With My Comment I really Like Paludis Too And I See You Are Already Doing Good Stuff For Paludis To Make It Still Works On Gentoo , Im Going To Clone Your Overlay And Give Paludis And Your Work a Try . I Don't Intended To offended You Sr.  But I Was With The Mentality Of IRC where you ask about "hey how i use paludis or hey i can install paludis" and right away is the answer We Dont Support Paludis ETC ETC , And The Poor Support In The #paludis Channel so hey excuse me please and keep doing what you do to keep paludis working on our distro and for real i would like to have paludis working on my new installation im doing now im recompiling the toolchain to start emerge -e world and would be nice to do it with paludis and yes portage is getting really slow and i need best and better faster pkg mangler too.

javashin:
Im Installing paludis-9999 from your forked overlay thanks for that @sandikata

sandikata:
I am asking there for a reason. I spend 1 day long to bump it, and make it work.
And it's fully working, if you do not use gentoo-lto, and maybe some bad overlays which using too old EAPI, paludis doing well with EAPI 7.
Exherbo team is not friendly, they are ex gentoo developers, but they're going rude, when you ask them for info about porting on gentoo. I understand them, Gentoo is terrible last years, and i preffer Funtoo, but funtoo do not have systemd, which is required for most of my everyday use.

sandikata:
and yeah, pkgcore is fastest gentoo package manager, but lack support for EAPI newer than 5 (EAPI 5 even still experimental)

javashin:
thanks for the info @sandikata  maybe @InBetweenNames  Can Make a Supported Stuff for Paludis I know This For Sure Paludis Is Fast Very Fast

InBetweenNames:
Yeah, I would like to support pkgcore as well.  There's no reason GentooLTO shouldn't be able to play nice with other similar package managers.

sandikata:
pkgcore is easy to support, its absolutely same syntax as in portage, just is fastest gentoo package manager, but unfortunately only experimental EAPI 5

javashin:
hey @sandikata paludis only can found python when boost is downgraded to version 1.65 and i had 1.69.0
there is no way to make it work with new boost ? and for what paludis need python ?

javashin:
paludis cant do what portage does i remember why i stop using it after emerge it configure it and when i ran cave resolve -c world to update a lot of stuff i set on /portage/profile is gone and wants to re emerge everything . cant and dont know about : like this one
i cannot mask use flags or force


no package.use.force 
root #echo 'sys-fs/zfs -kernel-builtin' >> /etc/portage/profile/package.use.mask
root #echo 'sys-fs/zfs kernel-builtin' >> /etc/portage/package.use/zfs.conf

sandikata:
you can everything.
But your approach is wrong. BTW, you can talk to me on FreeNode, i saw you in #bedrock channel. My nick is ROKO__
Paludis releated configs are stored in /etc/paludis. You can still use portage backward compatibility in paludis, but is known, as slow and unreliable.

Titaniumtown:
this was a trip to read

WillPower3309:
I think my PR #969 might fix this. @sandikata I would encourage you to check it out, as it might fix your issue without breaking any setups

Issue #307, opened 2019-05-09
New GentooLTO layout


InBetweenNames:
Hello everyone,

Just wanted to run an idea by you.  Currently, we symlink `make.conf.lto` and `make.conf.lto.defines` to `/etc/portage`.  This makes it hard to extend with different configurations.  What I'd like to do instead is only install things to `package.cflags` and `bashrc.d` and leave the rest of `/etc/portage` untouched.

One idea I had was to have users source directly from the overlay.  For example, instead of:

~~~
source make.conf.lto
~~~

One could do:

~~~
source /path/to/lto-overlay/sys-config/ltoize/files/configs/make.conf.lto
~~

It's longer, but this way we could more easily introduce more default configurations, such as:

* GentooLTO Experimental - `LTO, -O3, Graphite, -fipa-pta, -fno-plt, -fno-common`...
* GentooLTO Testing - `LTO, -O3`
* GentooLTO Stable - `LTO`
etc

The layout I'm proposing is:

* `lto-overlay/sys-config/ltoize/files/configs/*.lto` for default configurations
* `lto-overlay/sys-config/ltoize/files/configs/make.conf.lto.defines` for definitions

For a migration, we could provide a transition period where the symlinks are still installed, then afterwards we could remove them.  An `eselect news` entry could be made to warn users of the change.

I'd appreciate your thoughts on the matter!

Thanks,
Shane

javashin:
Whats The Flags Recommended For Speed/Stability Configuration ? not yet . Can  Someone Recommend A Standard Group Of *FLAGS That We Can Make Default For Future Reference Is a Lot To Process Reading All The Issues Looking For The Right Combination . Not Just For Me For Anyone new on The Overlay There Are A lot Choose ending With Smash Potato Gentoo Thanks.

gcs-github:
*Speaking only for myself:*

In order not to get surprises on my systems in case changes are made in the make.conf configuration, I don't source it at all, and instead choose to track the changes in this repository as well as copy-pasting what I find of interest.

Therefore, I wouldn't be affected by such a change.

For context, I find this combination of no sourcing + tracking changes to this repo to be a good way to stay up-to-date personally with what's going on here, plus learn as soon as new ideas are introduced when I don't already know about them.

pchome:
[...]

Issue #248, opened 2019-02-16
-fno-common workarounds


jelinekto:
I rebuilt my system couple weeks ago with `${NOCOMMON} ${FASTMATH}` enabled globally and thought I'd share list of packages that don't build with `-fno-common`. Note that the comment inside `make.conf.lto.defines` did not lie and there's quite a few breakages indeed. Probably isn't worth enabling as of now but in case someone's interested...

```
<x11-libs/gtk+-3 *FLAGS-="${NOCOMMON}"
app-arch/cpio *FLAGS-="${NOCOMMON}"
app-crypt/gnupg *FLAGS-="${NOCOMMON}"
app-editors/neovim *FLAGS-="${NOCOMMON}"
app-text/texlive-core *FLAGS-="${NOCOMMON}"
app-text/ttf2pk2 *FLAGS-="${NOCOMMON}"
dev-lang/R *FLAGS-="${NOCOMMON}"
dev-lang/orc *FLAGS-="${NOCOMMON}"
dev-libs/dbus-glib *FLAGS-="${NOCOMMON}"
dev-libs/ffcall *FLAGS-="${NOCOMMON}"
dev-libs/fribidi *FLAGS-="${NOCOMMON}"
dev-libs/gobject-introspection *FLAGS-="${NOCOMMON}"
dev-libs/libltdl *FLAGS-="${NOCOMMON}" # builds but causes symbol lookup errors
dev-libs/libmspack *FLAGS-="${NOCOMMON}"
dev-libs/libpipeline *FLAGS-="${NOCOMMON}"
dev-python/pygobject *FLAGS-="${NOCOMMON}"
gnome-base/librsvg *FLAGS-="${NOCOMMON}"
media-gfx/imagemagick *FLAGS-="${NOCOMMON}"
media-libs/freeglut *FLAGS-="${NOCOMMON}"
media-libs/gegl *FLAGS-="${NOCOMMON}"
media-libs/glu *FLAGS-="${NOCOMMON}"
media-libs/gstreamer *FLAGS-="${NOCOMMON}"
media-libs/mesa *FLAGS-="${NOCOMMON}"
media-libs/openal *FLAGS-="${NOCOMMON}"
media-sound/alsa-utils *FLAGS-="${NOCOMMON}"
media-sound/sox *FLAGS-="${NOCOMMON}"
media-sound/wavpack *FLAGS-="${NOCOMMON}"
[...]

PR #321, opened 2019-05-16
Include -fno-common workarounds

jelinekto:
After upgrading GCC to 9.1.0 I revised my ${NOCOMMON} workarounds I mentioned in #248 and since I noticed in #307 you are interested in introducing the flag, I thought I'd open a PR.

These do not build for me (save for one) when using `CFLAGS="-march=native -O3 -fgraphite-identity -floop-nest-optimize -fipa-pta -fno-math-errno -fno-trapping-math -fno-common -fno-semantic-interposition -flto=4 -fno-plt -falign-functions=32 -malign-data=cacheline -g0 -pipe"` globally.


InBetweenNames:
Excellent.  Thank you!  After the refactor I'd like to support `-fno-common` for sure.

Issue #164, opened 2018-11-02
Integration of Clear Linux patches


InBetweenNames:
Clear Linux maintains a number of performance related patches for open source projects.  There are quite a few: https://github.com/clearlinux-pkgs

It would be interesting to integrate these into GentooLTO somehow, either synced into the overlay or through a user install mechanism.

aw1cks:
I didn't look thoroughly, but if I'm seeing it right, these are just patches in  their git repo? It would be very easy to put the patches into /etc/portage/patches, is there some downside to this?

InBetweenNames:
It appears that way, however I'm guessing they also fine tune compiler options as well.  Also, I'm unsure if they use `icc` or `gcc` for their builds.  I'm guessing it depends on the package.  At the very least, we could include the source code patches, I think.

InBetweenNames:
Example: https://github.com/clearlinux-pkgs/fftw/blob/master/fftw.spec

InBetweenNames:
Interesting flag: `-fno-semantic-interposition`.  I'll be adding this one to the defaults I think.

InBetweenNames:
Other interesting things: they appear to enable `-ffast-math` on certain packages, or flags that `-ffast-math` enables at the very least.  Also interesting: they force all functions to be aligned on 32 byte boundaries with `-falign-functions=32`.  I wonder if they do that for AVX512 compatibility?  It'd be interesting to see what the benefits are for overaligning functions, if any.

InBetweenNames:
It appears that `-falign-functions=32` has some kind of impact on autovectorization.  By default on `x86_64`, it is set to `16`.

InBetweenNames:
I just tested an architecture that has AVX512 instructions:

~~~
> gcc -march=skylake-avx512 -flto -Ofast -Q --help=optimizer | grep falign-functions
  -falign-functions                     [disabled]
  -falign-functions=                    16
~~~

It seems by default on `x86_64`, `-falign-functions` is always set `16` (as long as `-O2` or higher is specified).

Althorion:
Phoronix [claims](https://www.phoronix.com/scan.php?page=article&item=ubuntu1810-fast-clear&num=1) that they build using GCC/Clang (differs from package to package).

InBetweenNames:
[...]

PR #323, opened 2019-05-18
sys-fs/btrfs-progs: disable gnu2 TLS dialect

jelinekto:
The `TLS relocation against invalid instruction` error as previously mentioned in #164 is still present.


InBetweenNames:
Thanks!

PR #332, opened 2019-05-26
dev-lang/python: Fixes emerge on Ofast systems

### LINKED ISSUE #200: python-2.7.15-r108[pgo] fails at tests
The new version of python 2.7 seems to fail to emerge due to failures in the instrumentation tests. I even used my fallback *FLAGS and the failures still occur. See:
[emerge_info.txt](https://github.com/InBetweenNames/gentooLTO/files/2597988/emerge_info.txt)
[build.log](https://github.com/InBetweenNames/gentooLTO/files/2597987/build.log)

The rest of my system is built with Ofast and fast-math, but I found it pretty odd, since the tests complain about functions that are implemented in python, such as [gamma](https://github.com/python/cpython/blob/master/Modules/mathmodule.c#L265) and [lgamma](https://github.com/python/cpython/blob/master/Modules/mathmodule.c#L366). What's more, I noticed the very same tests failing in the older version of the ebuild (python-2.7.15-r1) but there the emerge continues successfully.

### LINKED ISSUE #326: dev-lang/python hangs when using -Ofast
python-3.7.3 hangs during emerge when compiling test_buffers.py with -Ofast

asaparov:
Added -fno-finite-math-only workaround for all python builds (fixes #326). Also fixed an issue in the build for python-2.7.15-r108 where failing instrumentation tests would cause the entire emerge to fail, which is not the case for the other python versions (fixes #200).

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -125,6 +125,10 @@ media-libs/flac *FLAGS+=-fno-ipa-cp-clone # -fipa-cp-clone is enabled by default
 media-libs/faad2 *FLAGS+=-fno-tree-loop-vectorize # causes subtly wrong decoding
 # END: Deliberate -O3 workarounds
 
+# BEGIN: -Ofast workarounds
+dev-lang/python C*FLAGS+='-fno-finite-math-only' # instrumentation tests hang/segfault during emerge
+# END: -Ofast workarounds
+
 # BEGIN: Will not build with ninja
 x11-misc/polybar CMAKE_MAKEFILE_GENERATOR=emake
 app-doc/doxygen CMAKE_MAKEFILE_GENERATOR=emake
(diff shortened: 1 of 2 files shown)


sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
```diff
@@ -125,6 +125,10 @@ media-libs/flac *FLAGS+=-fno-ipa-cp-clone # -fipa-cp-clone is enabled by default
 media-libs/faad2 *FLAGS+=-fno-tree-loop-vectorize # causes subtly wrong decoding
 # END: Deliberate -O3 workarounds
 
+# BEGIN: -Ofast workarounds
+dev-lang/python C*FLAGS+='-fno-finite-math-only' # instrumentation tests hang/segfault during emerge
```
InBetweenNames: Just a clarification -- is C*FLAGS necessary or will *FLAGS work too?

asaparov: oh *FLAGS will work too

asaparov: should i change it to that?

glitchking007:
I can confirm my problem is resolved with these changes.

sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
InBetweenNames: If you don't mind -- it's just to keep things in line with the rest of the workarounds so that later it doesn't confuse anyone.

asaparov: sure no worries

InBetweenNames:
Thank you!

PR #334, opened 2019-05-28
Added handful of Ofast workarounds

asaparov:
Added the workarounds I use on my Ofast machines (amd64 KDE plasma). I expect we are missing a bunch from other desktop environments/architectures/etc, but hopefully these can get us started.

### DIFF
--- a/sys-config/ltoize/files/make.conf.lto.defines
+++ b/sys-config/ltoize/files/make.conf.lto.defines
@@ -43,7 +43,9 @@ NOCOMMON="-fno-common"
 # This option only affects C code.  Only non-conformant C code needs -fcommon, which is enabled by default.  Clear Linux leaves this flag off by default.
 # We may enable this at some point.  For now, there's a lot of breakages to work through.
 
-SAFERFASTMATH="-fno-math-errno -fno-trapping-math"
+SAFER_FAST_MATH="-fno-math-errno -fno-trapping-math"
+SAFE_UNSAFE_MATH_OPTS="-fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math"
+SAFE_FAST_MATH="${SAFE_UNSAFE_MATH_OPTS} -fno-math-errno -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fcx-limited-range -fexcess-precision=fast"
 
 #These are flags left off by default that we're planning to start using.  Clear Linux uses these in lieu of full -ffast-math optimizations
 #They DO break compliance with ISO C++, so we'll be careful about introducing these.
--- a/sys-config/ltoize/files/make.conf.lto.defines
+++ b/sys-config/ltoize/files/make.conf.lto.defines
@@ -43,9 +43,9 @@ NOCOMMON="-fno-common"
 # This option only affects C code.  Only non-conformant C code needs -fcommon, which is enabled by default.  Clear Linux leaves this flag off by default.
 # We may enable this at some point.  For now, there's a lot of breakages to work through.
 
-SAFER_FAST_MATH="-fno-math-errno -fno-trapping-math"
-SAFE_UNSAFE_MATH_OPTS="-fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math"
-SAFE_FAST_MATH="${SAFE_UNSAFE_MATH_OPTS} -fno-math-errno -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fcx-limited-range -fexcess-precision=fast"
+SAFEST_FAST_MATH="-fno-math-errno -fno-trapping-math"
+SAFER_UNSAFE_MATH_OPTS="-fno-signed-zeros -fno-trapping-math -fassociative-math -freciprocal-math"
+SAFER_FAST_MATH="${SAFER_UNSAFE_MATH_OPTS} -fno-math-errno -ffinite-math-only -fno-rounding-math -fno-signaling-nans -fcx-limited-range -fexcess-precision=fast"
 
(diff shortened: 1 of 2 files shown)


InBetweenNames:
Nice!  I'd really like to merge this.  Just to be clear: it seems that `SAFE_FAST_MATH` enables all `-Ofast` flags without actually defining `__FAST_MATH__` to get around build systems that try to prevent using it, is that correct?

asaparov:
Yes, so `kjs` and `opus` were the only packages on my machines that specifically check for `__FAST_MATH__`. I don't think I'm using anything on my system that can directly test `kjs`, but I'm not running into issues related to it. For `opus`, I can test it with examples [here](http://opus-codec.org/examples/) and they sound normal to me.

I noticed opus only checks for `__FAST_MATH__` because [they want to test whether a number is NaN by comparing it with itself](https://github.com/cisco/opus/blob/master/celt/arch.h#L154) `((x)!=(x))`, and `SAFE_FAST_MATH` would cause this to fail. A better solution for them would be to just use `isnan` from the standard library, which works at any optimization level.

InBetweenNames:
Yup that makes sense.  My only real comment is just the names of the variables -- is there any way to make them more descriptive to reflect their use case?  That way if another user comes along with the same problems, they might know what to try first.

asaparov:
Hmm, I'm not sure. From what I can tell, the flag `-funsafe-math-optimizations` does more than just enable `-fno-signed-zeros`, `-fno-trapping-math`, `-fassociative-math`, and `-freciprocal-math`. It also seems to affect behavior at link time. So SAFE_UNSAFE_MATH_OPTS tries to workaround this by enabling those four flags without the extra stuff that `-funsafe-math-optimizations` does.

SAFE_FAST_MATH enables all the other `-ffast-math` flags but also uses SAFE_UNSAFE_MATH_OPTS to avoid the extra effects of `-funsafe-math-optimizations`. But SAFE_FAST_MATH also avoids setting `__FAST_MATH__` so it can workaround compile-time checks.

Do you have any suggestions for better names?

InBetweenNames:
Perhaps a grading system of fast math might work.  I'm not sure what a good way to convey that via a variable name might be, but that way, a user could glance at the definitions in use and choose an appropriate one.  For example, `SAFEST_FAST_MATH` for just `-fno-math-errno -fno-trapping-math` and `SAFER_FAST_MATH` for enabling everything except `-funsafe-math-optimizations`.  I'm unsure if "safer/safest" are the best words, since I mean, it is a violation of the C and C++ specs, but I suppose it could work.  So, someone encountering a problem could first try `SAFER`, then try `SAFEST`, then try just plain old `-O3` (which would be like `NOFAST`).  What do you think?

asaparov:
I think that makes sense. I'll go ahead and make those changes. Though perhaps the first thing someone should try to debug `-Ofast`/`-ffast-math` is to just add `-fno-finite-math-only`. This works for a bunch of packages like `python`, `rrdtool`, `R`. Though to be honest, I don't know how common this will be since we have a total of only 6 `-Ofast` workarounds, so it's difficult to say what the best rule of thumb should be.

jelinekto:
Out of the default stage3 packages I can't build `sys-apps/coreutils` and `sys-apps/groff` with full `-Ofast`. `sys-apps/coreutils` builds with `-fno-finite-math-only` and for `sys-apps/groff` I enabled `-fno-unsafe-math-optimizations` for now.
@asaparov Can you reproduce these?

asaparov:
Ah interesting `coreutils-8.30` builds fine for me (which is what I had installed) but `8.31` fails with `multiple definition of 'minus_zero'` during linking. `-fno-finite-math-only` fixes this.

Same with `groff`, I had `1.22.3` installed but `1.22.4` doesn't build. `-fsigned-zeros` fixes this for me.

I'll add these workarounds to this pull request.

jelinekto:
I imagine having a look at [CloverOS' workarounds](https://gitgud.io/cloveros/cloveros/blob/master/binhost_settings/etc/portage/package.env) could be of some use.

So far I've ran into `x11-misc/redshift` malfunctioning which seems to be remedied by `-fno-finite-math-only`.

asaparov:
Oh thanks! This is useful. Though I noticed a number of packages on that list that build and work fine with `-Ofast` at least on my system: `dev-lang/spidermonkey`, `net-libs/nodejs`, `app-arch/bzip2`, `sys-boot/grub`, `dev-db/sqlite`, `app-arch/p7zip`, `sys-libs/db`, `sys-fs/ntfs3g`, `x11-libs/xcb-util-cursor`, `media-libs/glu`, `sys-libs/cracklib`, `media-libs/libjpeg-turbo`, `dev-lang/python-exec`, `app-arch/xz-utils`, `media-libs/libvpx`, and `app-crypt/gnupg`.

For `x11-misc/redshift`, it compiles fine on my system but it seems `-ffinite-math-only` causes a runtime issue where brightness values are interpreted incorrectly. I'll add a workaround for this.

I think since most of the workarounds in CloverOS are no longer needed, we should only add them as-needed. Also, the CloverOS workarounds seem to fall back to `-O3` which disables all of the `-ffast-math` optimizations, possibly unnecessarily. Hopefully as more people use `-Ofast` we can get wider coverage across more packages and environments.

jelinekto:
Interestingly, `=media-libs/opus-1.3.1-r1` compiles fine with full `-Ofast`.

asaparov:
Oh interesting, it seems `1.3.1-r1` ebuild specifically checks if `-ffast-math` or `-Ofast` is enabled, and if so, defines the `FLOAT_APPROX` preprocessor flag, which allows `opus` to compile with `__FAST_MATH__`.

I'll change the workaround to restrict it to older versions.

InBetweenNames:
Sorry for the delay on this one!  Thank you very much!

PR #308, opened 2019-05-10
media-libs/libwebp: gcc 9 fixed runtime issues caused by fipa-pta

asaparov:
Removed the workaround for this package. Is there a fixed section in the workarounds file?

Issue #257, opened 2019-03-12
-fipa-pta is universally broken in all GCC versions after 5.5.0 and should be disabled


Muffindrake:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88936

The bug remains `ASSIGNED` despite claims in one of the current PRs to this repository that `GCC 8.3.0 contained fixes`, despite no explicit mention of this bug in the 8.3.0 changelog.

InBetweenNames:
Wow, great find!  I've disabled `-fipa-pta` globally in this repo until this gets resolved.  Anyone using a supported configuration should receive the update to their `CFLAGS` when they next `emerge --sync`.  Hopefully this will be resolved in the next major GCC release.  I may post an eselect news at some point advising anyone using this flag to do a full system rebuild.

ionenwks:
I'm not surprised to see this confirmed. Had a lot of strange issues with it myself and had stopped using it pretty early rather than setting exceptions, often the problems weren't segfaults but strange behavior that were hard to diagnose later on (never safe to mess with that but I tried to build glibc with it once anyway, and then tried to rebuild gcc and it started giving errors with includes that felt completely unrelated)

barolo:
I've been using -fipa-pta on world since forever and had no issues whatsoever. I'm on AMD, is it confined to Intel only?

gcs-github:
Well, the problem is that `-fipa-pta` generates invalid code in specific situations. There's no guarantee that you would hit these cases at runtime, which can explain why you didn't notice any issue. I haven't had issues with the flag either (excepted maybe as properly handled by the LTO workarounds from this overlay) but the threat of invalid code generation is scary enough to warrant avoiding the flag by default.

This isn't specific to Intel, and an example of the problem with a reproducer is given in the GCC bug tracker link posted in this issue.

InBetweenNames:
Better to stay safe than sorry for sure.  I'll re-enable this once it's fixed upstream.

justanerd:
Now fixed in trunk:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=88936

InBetweenNames:
GCC 9.1.0 is now live and GentooLTO now supports it.  I'm leaving `-fipa-pta` out of the defaults for a few weeks while everyone upgrades.  However, interested users can opt in manually in their `CFLAGS`.  Leaving this issue open until it is re-added, in case problems arise.

InBetweenNames:
[...]

PR #340, opened 2019-06-02
Remove redundant -fipa-pta workarounds

jelinekto:
`media-sound/mpc` and `sys-devel/gdb` now work fine for me after #257 got fixed in GCC 9.1.

`media-libs/libwebp` appears to be ok with `-fipa-pta` too on my machine but I left that one in since the issue still might be present under some circumstances according to #308 . I can't reproduce that though, despite building with `-Ofast` myself.


InBetweenNames:
Does gdb attach segfault for you with -fipa-pta?

jelinekto:
I don't get any segfaults with neither stable nor testing version of gdb.

InBetweenNames:
Oops -- had an unstable overclock which was the culprit.  These microcode updates, I tell you!  Thank you for this, and I will test libwebp again shortly.

PR #341, opened 2019-06-02
Set LTO_ENABLE_FLAGOMATIC=yes for ebuilds providing USE='custom-cflags custom-optimization'

jelinekto:
Since we already have an override of this kind for `www-client/firefox`, I believe it makes sense to do the same for all the other ebuilds that make use of `custom-cflags` and `custom-optimization` flags.


InBetweenNames:
Looks good to me!  In fact, I will be adding in `custom-cflags` and `custom-optimization` into my `USE` from now on.  Thanks!

Issue #358, opened 2019-06-21
dev-lang/php segfaults with -fipa-pta


lucius-martius:
I have a small server running nextcloud-16.0.1 + mariadb-10.3.15 + apache-2.4.39 + php-7.3.6.

When I start the server it looks like it's running, but the apache2 child processes continuously segfault.

The log contains messages like this:
[Fri Jun 21 02:01:54.615603 2019] [core:notice] [pid 27274] AH00052: child pid 19517 exit signal Segmentation fault (11)

When attaching gdb to a running process, the crash can be traced back to the function  zend_get_executed_filename().

As a workaround i can just disable -fipa-pta for php and the segfaults stop:
dev-lang/php *FLAGS-="-fipa-pta"

I had already tried a full world rebuild.

Here is my emerge --info: [emerge-info.txt](https://github.com/InBetweenNames/gentooLTO/files/3312542/emerge-info.txt)

InBetweenNames:
Good find!  Would you be willing to do a PR for this?

PR #359, opened 2019-06-21
dev-lang/php: Disable -fipa-pta

lucius-martius:
When compiled with the flag, PHP was causing apache to segfault under certain conditions.

Also see issue #358 for more information.


InBetweenNames:
Thanks!

PR #370, opened 2019-07-04
Add perliminary Serbo-Croatian translations and some README.md tweaks

elsandosgrande:
It is as the title says.

The translations are preliminary, but could encourage others to translate into their languages as well. Obviously, I shall complete the translations, but not within one day, especially when typing two documents out. 

The tweaks, as can be seen in the commits, almost exclusively concern markdown adjustments and text formatting oddities. I only changed "LTOed" to "LTO-ized" so that it may grammatically stay in accordance with the rest of the document (the package is named `ltoize`, so it would make sense for the past tense of the verb/package name to be, when adjusted for correct orthography, "LTO-ized").


InBetweenNames:
Thank you!

elsandosgrande:
You are welcome!

PR #388, opened 2019-07-23
sys-boot/systemd-boot: disable gnu2 TLS dialect

jelinekto:
Another `TLS relocation against invalid instruction` error.


elsandosgrande:
Interesting. I have no such issue with the `gnuefi` USE flag for systemd.

InBetweenNames:
Thank you!

elsandosgrande:
I just realized that it is for the standalone package.

PR #407, opened 2019-08-20
net-libs/nodejs: Added `-ffinite-math-only` workaround

asaparov:
Added workaround for `net-libs/nodejs`. `ffinite-math-only` (implied by `-Ofast` and `-ffast-math`) causes `npm` to fail on startup.


InBetweenNames:
Thank you!

Issue #403, opened 2019-08-14
sys-apps/man-db: whatis stops with SIGABRT at runtime when compiled with -flto


lucius-martius:
I have noticed this when using tab-completion with the fish shell. The command does what it should, but fails with SIGABRT, which leads to the tab-completion of fish to fail as well.

It can also be reproduced when calling apropos (/usr/bin/whatis) manually:

```
~> apropos apropos
apropos (1)          - search the manual page names and descriptions
free(): invalid pointer
fish: 'apropos apropos' terminated by signal SIGABRT (Abort)
```

This behaviour seems to be new with the testing versions of `>=sys-apps/man-db-2.8.5`,  as version `2.7.6.1-r2` does not show the issue.

A workaround I found is is to disable LTO with `sys-apps/man-db *FLAGS-=-flto*`. If others can confirm this issue I can create a pull request for this workaround.

javashin:
i know about this 
readline ncurses bash with no lto too

perfect7gentleman:
no error

InBetweenNames:
Confirmed!

PR #410, opened 2019-08-24
sys-apps/man-db: Disable LTO due to runtime issues (#403)

lucius-martius:
LTO disabled for sys-apps/man-db.

See #403 for more info.


InBetweenNames:

PR #422, opened 2019-10-11
sys-config/ltoize: make graphite a choice when using lto-overlay

jiblime:
Add optional graphite USE flag to pull in sys-devel/gcc[graphite] as a build dependency

Revision reason: Ebuilds that use GCC as their compiler or do not strip flags will cause the compiler fail when using `-fgraphite-identity -floop-nest-optimize` without building GCC with USE graphite or `--enable-isl`.

The added USE flag in ltoize only adds a forced reminder to build GCC with graphite. Users on a new system can emerge ltoize prior to re-emerging GCC for ltoworkarounds.conf / make.conf.lto CFLAGS without needing to use a workaround (although that would beat the purpose of it).

### DIFF
--- a/sys-config/ltoize/ltoize-0.9.4-r2.ebuild
+++ b/sys-config/ltoize/ltoize-0.9.4-r2.ebuild
@@ -1,0 +1,109 @@
+# Copyright 1999-2019 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit toolchain-funcs
+
+DESCRIPTION="A configuration for portage to make building with LTO easy."
+HOMEPAGE="https://github.com/InBetweenNames/gentooLTO"
+KEYWORDS="~amd64 ~x86 ~arm"
+
+SRC_URI=""
+
+LICENSE="GPL-2+"
+SLOT="0"
+IUSE="+graphite override-flagomatic"
+
+#portage-bashrc-mv can be obtained from mv overlay
+DEPEND="
+	>=sys-devel/binutils-2.32:*
+	>=sys-devel/gcc-config-1.9.1
+	|| (
+		>=sys-apps/portage-2.3.52
+		>=sys-apps/portage-mgorny-2.3.51.1
+	)
+	app-portage/portage-bashrc-mv[cflags]
+	"
+
+RDEPEND="${DEPEND}
+	graphite? ( >=sys-devel/gcc-4.8.5[graphite] )
+"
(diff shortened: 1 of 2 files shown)


wolfwood:
can you add a sed expression to remove ${GRAPHITE} from the CFLAGS line in make.conf.lto if the use flag isn't set? then this would work as @bakape expected, without user intervention.

jiblime:
Good idea, I've added that. I think it'd be cleaner to add to make.conf.lto.defines though, since some people have custom variables in make.conf.lto; it could accidentally remove ${GRAPHITE} from the wrong areas. Here's a different version that comments out GRAPHITE from make.conf.lto.defines, I'm in between on which is ideal.

https://github.com/jiblime/gentooLTO/commit/a289abea06167d97aea4f51bb69f8b3468c0bf06

wolfwood:
@jiblime my read is that since make.conf.lto is a symlink, the user won't have edits preserved?

removing it from defines is like deleting the definition so if the use does add, it already have graphit enabled, it's an extra hurdle to work around. i feel that we only want to remove the use of ${GRAPHITE}.

I do agree that the best solution would be if we could use bashrc.d but i wish it could check the gcc USE flags. i guess just dropping wild carded graphite flag removes in a no-graphite.cflags file would give the user a chance to delete the file wholesale when they address the problem.

jiblime:
@wolfwood, agreed on the removal from defines. I actually had pushed a commit to the wrong branch regarding using bashrc.d, let me change that.

On the symlink, I don't really understand what you mean...

wolfwood:
@jiblime what i meant was, aren't make.conf.lto and make.conf.lto.defines overwritten everything the ltoize pkg is updated?

jiblime:
@wolfwood, oh yeah, I usually stash or have a backup it when that happens, but I did overwrite it once...to your point I keep much of my edits to make.conf.lto, maybe I've been doing it wrong lol.

BTW, sorry for the git spam.

jiblime:
Update: I don't like the current implementation. Building Graphite support doesn't add an astronomical time to emerging so I think the USE should be forced, like `(graphite)`. If one chooses to go make their own workaround to build GCC without Graphite support, it would make sense to append ${GRAPHITE} flag filters to the `41-lto-patch.sh`, and to be silent about it to prevent polluting build logs.

wolfwood:
I support the (graphite) required USE that depends on gcc graphite

InBetweenNames:
Indeed, I think that's the way to go -- force `gcc[graphite]` but make Graphite an optional thing (i.e., take it out of the default GentooLTO configuration).

PR #430, opened 2019-10-27
dev-scheme/guile: disable -ffinite-math-only on -Ofast systems

jelinekto:
This works around `Floating point exception` error during the build.


InBetweenNames:
Thank you!

Issue #432, opened 2019-10-31
tensorflow-estimator can't find __cpu_model if tensorflow is built with LTO


wolfwood:

InBetweenNames:
Fixed by PR.  Thanks!

PR #433, opened 2019-10-31
sci-libs/tensorflow # Issue #432 tensorflow-estimator fails with missing symbol __cpu_model

wolfwood:
fix for #432


InBetweenNames:
Thank you!

Issue #450, opened 2019-12-06
Fortran GRAPHITE ICE with scipy-1.1.0


wolfwood:
resolved by adding `sci-libs/scipy CFLAGS-="${GRAPHITE}"`

However, since the ICE is in the fortran compiler, but this package also contains C code, it would be nice to be able to turn off GRAPHITE for F*FLAGS only.  I also noticed that setting FFLAGS/FCFLAGS/F77FLAGS in my make.conf is ignored and they just get set to `${CFLAGS}`

Before I ask @vaeth to enhance portage-bashrc-mv to add F*FLAGS, curious if anyone else can reproduce this issue?

wolfwood:
[scipy-1.1.0-build.log.txt](https://github.com/InBetweenNames/gentooLTO/files/3934209/scipy-1.1.0-build.log.txt)

wolfwood:
https://bugs.gentoo.org/575446 pertains to setting F*FLAGS in cmake.conf

wolfwood:
this is a reduction of the the offending code: [scipy-graphite-ice.f.txt](https://github.com/InBetweenNames/gentooLTO/files/3964654/scipy-graphite-ice.f.txt)

`gfortran -O2 -floop-nest-optimize scipy-graphite-ice.f`
give me 
```
wolfwood@Pixel ~ (Pixel⚡) $ gfortran -O2 -floop-nest-optimize /tmp/scipy-grapite-ice.f
/tmp/scipy-grapite-ice.f:16:72:

   16 | 30            ZQ=ZQ+ZW
      |                                                                        1
Warning: Fortran 2018 deleted feature: DO termination statement which is not END DO or CONTINUE with label 30 at (1)
/tmp/scipy-grapite-ice.f:20:72:

   20 | 35         ZO(NR)=Z
      |                                                                        1
Warning: Fortran 2018 deleted feature: DO termination statement which is not END DO or CONTINUE with label 35 at (1)
during GIMPLE pass: graphite
/tmp/scipy-grapite-ice.f:1:0:

    1 |         SUBROUTINE CERZO(NT,ZO)
      | 
internal compiler error: Segmentation fault
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://bugs.gentoo.org/> for instructions.
```

@jiblime since you aren't seeing the dav1d issue #442, I'm curious if this reproduces for you or not?

PR #454, opened 2019-12-15
sci-libs/scipy: graphite ICE

wolfwood:
temporary workaround for #450

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -108,6 +108,7 @@ sci-libs/tensorflow *FLAGS-=-flto* # Issue #432 tensorflow-estimator fails with 
 #Packages which Graphite optimizations don't play nice with
 # BEGIN: Graphite ICE (Internal Compiler Error)
 # Report them to GCC team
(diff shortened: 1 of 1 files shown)


elsandosgrande:
@wolfwood Only removing the `-floop-nest-optimize` flag is necessary.

wolfwood:
@InBetweenNames this patch should also include modifying the ltoize ebuild dependency to >=app-portage/portage-bashrc-mv-19.0, as the F*FLAGS syntax was just introduced, but I wasn't sure if you'd want to do a revision bump or what.

InBetweenNames:
Ah yes, you're right.  After I'm done marking exams I'll make sure this and the graphite patch are merged, and the dependency on portage-bashrc-mv-19.0 or higher is added to ltoize.

wolfwood:
@elsandosgrande you are correct. patched.

wolfwood:
@InBetweenNames I figured it was that time of year; no rush.

elsandosgrande:
Also, [from what I know](https://github.blog/2013-01-22-closing-issues-via-commit-messages/), you can automatically close the aforementioned ticket if you add "Fixes #blahblah" or "Closes #blahblah" or a few other possible keywords to either a commit which is covered by the pull request or to the very pull request's description itself (this part is mostly me guessing, since I have only seen it done like this once).

wolfwood:
@InBetweenNames trofi just merged a pre-release patch for isl that fixes this issue today, lol.

Issue #442, opened 2019-11-17
media-libs/dav1d doesn't build with graphite


ElDavoo:
```
-O2 -fno-stack-protector -march=native -falign-functions=32 -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=1 -fuse-linker-plugin -Wl,-O1 -Wl,--as-needed -Wl,-O1 -Wl,--as-needed -O2 -fno-stack-protector -march=native -falign-functions=32 -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=1 -fuse-linker-plugin -pthread -Wl,--start-group -lm -Wl,--end-group
during GIMPLE pass: graphite
../dav1d-0.5.1/tests/checkasm/looprestoration.c: In function ‘check_sgr.isra.0’:
../dav1d-0.5.1/tests/checkasm/looprestoration.c:118:13: internal compiler error: Segmentation fault
  118 | static void check_sgr(Dav1dLoopRestorationDSPContext *const c) {
      |             ^
```
Disabling graphite fixes the issue and makes dav1d compile correctly.

jiblime:
I have had no troubles compiling with bfd nor gold.

```
=================================================================
                        Package Settings
=================================================================

media-libs/dav1d-0.5.1::gentoo was built with the following:
USE="10bit 8bit asm" ABI_X86="(64) -32 (-x32)"
CFLAGS="-O3 -march=native -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fno-plt -fipa-pta -fno-semantic-interposition -fno-math-errno -fno-trapping-math -flto=12 -malign-data=cacheline -pipe -fuse-ld=bfd -falign-functions=32 -Wl,-O1 -Wl,--as-needed"
CXXFLAGS="-O3 -march=native -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fno-plt -fipa-pta -fno-semantic-interposition -fno-math-errno -fno-trapping-math -flto=12 -malign-data=cacheline -pipe -fuse-ld=bfd -falign-functions=32 -Wl,-O1 -Wl,--as-needed"
[...]

PR #455, opened 2019-12-15
media-libs/dav1d: workaround for graphite ICE, issue #442

wolfwood:
disable test to workaround #442, patch courtesy of @pchome


InBetweenNames:
Thank you!

PR #457, opened 2019-12-24
app-emulation/libpod: Disable LTO

aidanharris:
Versions 1.6.3-r2 and 1.6.4 do not play nice with LTO.

This is a golang application, so this is probably expected (its linker complained about the LTO flags)?


InBetweenNames:
Thank you!

Issue #440, opened 2019-11-15
dev-lang/python-3.8.0 fails to build with LTO


ElDavoo:
See  https://bugs.gentoo.org/700012 , explains much better than I do.
Removing -flto fixes the issue for me.

ElDavoo:
It also looks like fat objects can fix the issue too.

elsandosgrande:
I can confirm that adding the fat allows it to compile again.

pchome:
@elsandosgrande 
> I can confirm that adding the fat allows it to compile again.

```
    IN CASE OF
        LTO
-------------------
        USE
 -ffat-lto-objects
```

TL;DR: use `ax_cv_c_float_words_bigendian=no` on "regular" PC system.

Problem source: https://www.gnu.org/software/autoconf-archive/ax_c_float_words_bigendian.html

@InBetweenNames, to solve many of this `autotools` problems (including "variable substitution" problem) it maybe worth to create the collection of such problematic tests and generate `make.conf.lto.autoconf`, which will contain `_cv_` variables.
I guess it can be separate project w/ stripped `-flto` in ebuild, or `configure.in` as part of `sys-config/ltoize`. Not so perfect solution, but should fix many of current and future problems (and shrink the `ltoworkarounds.conf` a bit, I guess).

InBetweenNames:
@pchome Agreed -- I have had similar thoughts too.  Most of our problems at this point are the configure scripts not playing well with LTO.  I only wish I had more time to tackle it!

InBetweenNames:
Was able to bypass this by setting it manually as the config stated.

InBetweenNames:
Sigh.  Yet another autoconf macro that depends on compiler internals.  So essentially, the test string in the macro isn't present in LTO object file output, so it is unable to determine the endianness of floats in the system that are larger than one word.

For convenience, we can build with fat lto objects -- the result should be identical.  For users that want to really be sure only LTO is being used, edit `configure` manually with the correct value.  It is most likely `no`:

~~~
14250c14250
< ax_cv_c_float_words_bigendian=unknown
---
> ax_cv_c_float_words_bigendian=no
~~~

I would include a patch to do this, but I am aware that some users may be running GentooLTO on systems where the answer would be `yes` and I don't want to cause problems for those users.

You can include this patch in your `/etc/portage/patches/` if needed.

jiblime:
[...]

PR #460, opened 2020-01-01
>=dev-lang/python-3.8: Build with fat LTO

2l42h3r:
Will fail during configure otherwise.


elsandosgrande:
@2l42h3r Why did you close it all of a sudden?

2l42h3r:
Seen this was already talked about in [this issue](https://github.com/InBetweenNames/gentooLTO/issues/440), but Shane said including a patch for the configure script is a better option.

elsandosgrande:
Oh, all right. I still don't understand how some people can just close their pull requests or tickets without providing context or reason when there isn't any obvious context or reason to do so (at least obvious at first, but it's still a guessing game for me at best).

PR #466, opened 2020-01-17
Fix for Undefined symbol in x11-base/xorg-server

ElDavoo:
This change fixes this error:
Failed to load /usr/lib64/xorg/modules/drivers/modesetting_drv.so: /usr/lib64/xorg/modules/drivers/modesetting_drv.so: undefined symbol: glamor_block_handler
When trying to start x11-base/xorg-server-1.20.7 with an i965 video card.

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -236,6 +236,7 @@ dev-libs/nss *FLAGS-="-fno-plt"
 sys-libs/glibc *FLAGS-="-fno-plt" #When using prelink, will cause a bunch of segfaults.
 media-video/ffmpeg *FLAGS-="-fno-plt"
 app-emulation/wine-staging *FLAGS-="-fno-plt"
+app-emulation/wine-vanilla *FLAGS-="-fno-plt"
 dev-db/mysql-connector-c *FLAGS-="-fno-plt" # SIGABRT during compile if not used
 sys-libs/compiler-rt-sanitizers *FLAGS-="-fno-plt"
 x11-base/xorg-server *FLAGS-="-fno-plt" # Undefined symbol in i956 driver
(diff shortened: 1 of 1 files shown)


elsandosgrande:
I would say that this should be restricted to the i965 extended USE flag until somebody with a different driver complains about the same thing.

InBetweenNames:
Actually, we do have precedent for that in `ltoworkarounds.conf`, but it looks like it might be a bit trickier as `xorg-server` itself doesn't directly use the `VIDEO_CARDS` variable.  It would be nice to do something like the following:

`x11-base/xorg-server  "has video_cards_i965 ${IUSE//+} && use video_cards_i965 && FlagSubAllFlags -fno-plt" # Undefined symbol in i956 driver`

elsandosgrande:
OK, where did you learn this Bash trickery?! I have no idea how to even look this up on DuckDuckGo.

pchome:
@elsandosgrande 
Probably my fault, I cursed them. ;)
Based on **mv** and **eclass** sources.

elsandosgrande:
Hm... @pchome 

1. Where can I find further reading on this?
2. Thank you for the explanation (although I am still a bit fuzzy on `${IUSE//+}`)!
3. Let me guess, this will not work for ABI and/or CHOST environment variables, will it? I am asking because there seem to be a couple of flags that only cause problems on ARM (there is a ticket open here, but I do not have the link at hand right now and am in the middle of a very late dinner, so just look for Jannik[some numbers which I do not recall]).

Something tells me that we should probably move this particular discussion onto the Gitter group, given that it is starting to drift off of the pull request at hand (I am talking about my questions above this paragraph).

pchome:
1. by reading **eclass** and **mv** sources, obviously
2. some defaulted +flag,  removes **+**
3. dunno, this was introduced as single-time fix for Wine custom flags IIRC, if you need advanced check -- just do advanced check.

No discussion here, I just clarified some moments, nothing to discuss. **Just to prevent you from doing wrong thing to repository in future.**

Jarel1337:
I would like that I am affected by this as well (running in a KVM/QEMU VM) and the fix proposed did fix it.

ElDavoo:
Please accept this pull request as I'm having this problem with recent updates too

InBetweenNames:
Alright, I'll accept this PR for now as it looks like you can't even access `USE_EXPAND` flags from the ebuild if they're not in `IUSE`.  It's like they get stripped out of the variable.  One day when I get time, I'll deep dive this and see what's happening under the hood.

AdelKS:
Hello everyone,

I hit this same issue independently recently, and went to open a bug recently: https://bugs.gentoo.org/778494
There's a better fix, that has to be done manually: modules in xorg need to be loaded in certain order and `-fno-plt` will work with xorg and its related drivers. More information available in the bug report.

Cheers!

Issue #453, opened 2019-12-10
sys-libs/libcap-2.28 failing to build with LTO


jelinekto:
I'm not able to build the latest `libcap` update even with plain `-march=native -O1 -flto -pipe`. Gold throws out an error (see the log), bfd build hangs indefinitely. `-ffat-lto-objects` seems to be of no use here.
Can someone reproduce this?
[build.log](https://github.com/InBetweenNames/gentooLTO/files/3944663/build.log)

corrmaan:
`=sys-libs/libcap-2.28::gentoo *FLAGS-=-flto*`

Works for me

jiblime:
Unable to replicate after emerging successfully, but I was getting stuck when running in parallel.

`MAKEOPTS="" emerge -1 libcap` fixed it for me.

elsandosgrande:
I am using the the BFD linker and I am not stuck, but getting the aforementioned undefined references. Only disabling LTO works for me. I have tried your workaround @jiblime, but it did not work for me. Yes, I am able to replicate it after emerging the package successfully. I event tried changing `NTHREADS` to 1, but to no avail.

jiblime:
https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=d6502cd58680cab758144bfe51ec6c5556179695

Maybe try resyncing your trees, failing that you can just dash the test as described here https://bugs.gentoo.org/702516#c9

elsandosgrande:
I'm currently in the middle of an update. I synced about ten hours ago.

jelinekto:
Okay, I am now able to build LTOed `sys-libs/libcap-2.28-r1` using `-fuse-ld=bfd`. Gold build still fails for me.

elsandosgrande:
Hm... This is not the first instance of Gold failing where BFD does not. That's actually why we now recommend BFD by default (read the updated README if you have not already).

wolfwood:
same issue with `__wrap_pthread_create` seems to be present with sys-auth/elogind. it looks like removing `-Wl,--as-needed` / adding `-Wl,--no-as-needed`to LDFLAGS is recommended for this kind of issue

wolfwood:
`LDFLAGS-="-Wl,--as-needed" LDFLAGS+="-Wl,--no-as-needed"` doesn't seem to work (probably needed for glibc, instead), but `sys-auth/elogind LDFLAGS+="-Wl,-fuse-ld=bfd"` does.

[...]

PR #468, opened 2020-01-18
sys-libs/libcap: use ld.bfd

jelinekto:
Issue #453

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -251,6 +251,7 @@ media-libs/libcaca *FLAGS+="-fuse-ld=bfd"
 dev-util/kbuild *FLAGS+="-fuse-ld=bfd"
 games-emulation/pcsx2 *FLAGS+="-fuse-ld=bfd"
 net-misc/curl *FLAGS+="-fuse-ld=bfd"
+>=sys-libs/libcap-2.28 *FLAGS+="-fuse-ld=bfd" # Issue #453
 # END: GOLD linker workarounds
 
 # BEGIN: -fno-common workarounds


elsandosgrande:
Maybe you should also add `*FLAGS-="-fuse-ld=gold"` in case somebody is using the same flag to enforce the usage of the Gold linker (I certainly can't be bothered to manage that through the `binutils` configuration).

InBetweenNames:
Thank you!

barolo:
@elsandosgrande You can't manage linkers through binutils-config any-longer it was removed very very recently

elsandosgrande:
@barolo 

1. How recently (like, one month, two months, ...)?
2. Huh, I guess I chose the correct method when I stated using GentooLTO...
3. If I may ask, why do you never appear on Gitter?

barolo:
@elsandosgrande
1. In last few updates, within last month, 5.2 no longer has that ability and I've seen post about it somewhere few days ago
2. Probably.
3. Do I have to? I'm not very social, it tires me easily. I just pop out from time to time.

elsandosgrande:
@barolo 

1. Go figure. Out of curiosity, were the posts on the forums, or...?
2. Yay me!
3. Well, something tells me that somebody (most likely Shane) will lose their patience with conversations like these being held in tickets and requests instead of on Gitter, the **chat** client. I mean, that's the whole reason why I suggested it in the first place over half a year ago if memory serves.

barolo:
@elsandosgrande 
1. https://forums.gentoo.org/viewtopic-t-1106116-start-0.html
3. Fine, seeing that gitter [ I know what it is ] have sensible irc bridges [ yeah I'm that guy ] I'll join

elsandosgrande:
Well... I'm seventeen, so...

Also, why use an IRC bridge when I see that you have already logged into Gitter in the past at one point (I'm guessing that that was part of you checking it out; you are listed as one of the members of the group)? I mean, IRC does not maintain chat history, so...

PR #469, opened 2020-01-18
Add -fno-common workarounds

jelinekto:


InBetweenNames:
Thank you!  It's looking like this list will prove especially handy for GCC 10 users, where `-fno-common` is default.

PR #471, opened 2020-01-19
Gcc needs to be compiled with graphite in USE

Bryce-MW:
It took me forever to figure out why gcc could not find isl even after installing different versions and recompiling gcc. I found by looking through solved issues that gcc needs to be compiled with graphite in the USE. Maybe it's obvious to more experienced people but I felt that something should be added specifying that this needs to be done since the error makes you go down the tree of thinking the issue is with isl. Feel free to reject this, I'm pretty new to Gentoo so I don't know what things people need to be told or not. It could also maybe be made a bit more obvious that there is a chat group. I didn't notice the glitter badge for quite a while. I guess I was just not observant enough so it's probably not needed.

Either way, now that I have this working, I am excited to try it out! I do a lot of networking and am considering witching some of my Linux routers to Gentoo so I will be able to test various networking packages which are probably rather obscure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ The biggest gotcha with `-O3` is that it does not play nice at all with Undefine
 
 ## How to use this configuration
 
-Add the `mv` overlay (`layman -a mv`) and then add this overlay (`layman -a lto-overlay`) to your system and run `emerge sys-config/ltoize`. Add the package to your `/etc/portage/package.accept_keywords` if necessary. This will add the necessary overrides to `/etc/portage/`, but it won't modify your `make.conf`. It will create a `make.conf.lto` symlink in `/etc/portage` with the default GentooLTO configuration. You may need to add `graphite` to your `USE` and recompile `gcc` before adding the GentooLTO configuration to your `make.conf` file. To use the default configuration, define a variable `NTHREADS` with the number of threads you want to use for LTO. Then, source the file in your own `make.conf` like in this example:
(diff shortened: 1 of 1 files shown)


elsandosgrande:
> ... add `graphite` to your `USE` and recompile `gcc`...

My suggestion is:

> ... add `graphite` to your USE flags and recompile GCC...

BryceAS202313:
> > ... add `graphite` to your `USE` and recompile `gcc`...
> 
> My suggestion is:
> 
> > ... add `graphite` to your USE flags and recompile GCC...

That sounds like a good idea, I've gone ahead and changed that.

elsandosgrande:
Also, regarding the Gitter badge (not glitter), there was one other person that missed it. While I (and probably a good number of people) notice the badges on a README before I read the README, these cases might just warrant a notice in the main body of the README. With that said, the other person in question missed a few other things that *were* in the main body (I had even made one of them bold when I did some touch-ups of the README), so...

InBetweenNames:
Thanks for this -- related PR #422

elsandosgrande:
Hm... So, I see two ways of going about this:

1. Waiting until the other pull request is accepted and then accordingly make changes in this one.
2. Accept this and then make those changes after the other pull request is accepted.

BryceAS202313:
I have just noticed that in make.conf.lto.defines, it actually does say that GCC needs to be compiled with the graphite USE flag. I didn't notice before because I was using the simple method (make.conf.lto) so I never thought to look in make.conf.lto.defines

I had a look at PR #422, I think that that should be accepted then slightly modify this one. Given that that one forces GCC to be compiled with graphite, we would only need to comment on how to disable it (/enable it if graphite is left off as default).

acjones8:
This is a little bit of an old request, but I just want to comment that I also ran into this exact same issue while switching to LTO. I didn't realize GCC isn't compiled with graphite support by default, and you have to go digging through a couple logs before it'll tell you that. I definitely think it would be a good idea to merge this and add the note to the README!

telans:
@InBetweenNames this really should be merged

PR #474, opened 2020-01-22
sci-misc/boinc: build without lto

Uroc327:
boinc_client crashes on startup due to a buffer overflow when built with -flto. Building with fat lto objects did not help on my machine. Other flags (graphite, no sem interpos, ipa, devirt) seem fine.


elsandosgrande:
What do you use the package for? I would like to try and recreate the issue on my laptop, just to be thorough. 

Also, while I do not know if Shane minds this at all, I think that opening a ticket before a pull request for a workaround and discussing the issue at hand (or, at least, having at least one more person recreate the issue in said ticket) would be good "etiquette", if one may call it that.

Uroc327:
boinc is a platform to 'donate' computing power to science. See [their site](https://boinc.berkeley.edu/).

Oh, sry. I am used to also discuss pull requests. But I guess it's nice to have an issue to discuss this and possibly track multiple pull requests. I'll do this the next time.

elsandosgrande:
Well, since we are already here... Do you know exactly what triggered the buffer overflow? I do not mean "Do you exactly know which piece of source code is causing this behavior when the program is compiled with link-time optimization?", but "What exactly(-ish) was the program doing before it failed due to a buffer overflow?". 

Also, it will be a few days before I can check the package out (rebuilding the world with pre-release GCC 10).

Uroc327:
Not really, no. Judging by the messages the program was initializing itself. This happened pretty much directly after calling `boinc_client`.

elsandosgrande:
Well... That's certainly lovely...

InBetweenNames:
Thank you!

PR #481, opened 2020-01-28
dev-python/numpy: Added `-ffinite-math-only` workaround

### LINKED ISSUE #470: -ffinite-math-only is bad for dev-python/numpy
see https://github.com/numpy/numpy/issues/15344

asaparov:
Added `-ffinite-math-only` workaround for `dev-python/numpy`, which fixes #470.


InBetweenNames:
Thanks!  Better be safe than sorry with this one.

PR #486, opened 2020-02-17
Changes for sys-libs/libapparmor and media-sound/musescore

Jannik2099:
Musescore does not compile at all. Libapparmor compiles, but causes undefined symbol errors when compiling sys-apps/apparmor

I forgot to save the build logs, I can reemerge these packages and submit them if needed


InBetweenNames:
Hi @Jannik2099 , would you mind creating a PR for just the `apparmor` workaround?  I'm pretty sure that the `-Wl,` bug for musescore isn't LTO related and is a package bug that should be fixed in the main tree.

Jannik2099:
Hey Shane, you're right, this is not related to lto and happens without lto aswell. I reverted the commit

Quick question: if I have multiple fixes that should get merged, do you prefer one PR for each or all combined into one, seperated as commits?

InBetweenNames:
Sorry about the delay -- I'm usually okay with a few workarounds bundled together so long as they are related.  I'm not a big fan of mega-PRs although I will review them and advise how to break them down into smaller ones.  Generally my rule is, if I can't review it within 10 seconds, it goes on the backburner until I get time to sift through and test each change.  That's just me being a lazy maintainer though :)

InBetweenNames:
Also, thank you for the contribution!

PR #502, opened 2020-04-14
sci-visualization/paraview: Change CFLAGS

corrmaan:
I was attempting to build the newest available paraview (5.8.0-r2 on gentoo's default repo) and was constantly running out of memory during the build. At around 91% of the way through my RAM and swap (32 GB + 40 GB) would be maxed out (-j16). Changing to -j2 still would not fix the issue. I did not have the patience for -j1.
So after enabling all the [debugging recommendations](https://github.com/InBetweenNames/gentooLTO/wiki/Debugging#workflow-for-debugging-a-build-failure) and disabling them one-by-one, starting with `*FLAGS-=-flto*`, then `/-O3/-O2` etc., the build could finish with the graphite flags disabled. I didn't go further to remove individual `${GRAPHITE}` flags (i.e. at the time of writing was `GRAPHITE="-fgraphite-identity -floop-nest-optimize"`). If someone has any insight in to the memory effects of either of these flags then let's please discuss it.


InBetweenNames:
Great, thanks for looking into this!  Would you mind just adding a comment after the entry explaining that removing `GRAPHITE` allows the build to complete?

corrmaan:
Sure thing, done!

InBetweenNames:
Thank you!

PR #505, opened 2020-04-19
app-portage/lto-rebuild: fix '-a' and '-l' not matching their help description.

TheGreatMcPain:
lto-rebuild currently will list archives when using '-l', and will list packages when using '-a'.
This doesn't match the help description.


InBetweenNames:
Looks good to me!  Would you be alright with also including a version bump on `app-portage/lto-rebuild` to `0.9.8` in this PR?

TheGreatMcPain:
Yup, got it done.

InBetweenNames:

PR #512, opened 2020-04-29
dev-python/pillow: added `-fno-finite-math-only` workaround.

asaparov:
Although it compiles fine without the workaround, it causes `import matplotlib.pyplot` to fail with `undefined symbol: __log_finite`.


InBetweenNames:
Thank you!

PR #515, opened 2020-05-08
sys-devel/llvm: added `-fno-finite-math-only` workaround

asaparov:
Without the workaround, llvm would compile fine, but clang-10 would fail to emerge with error ``undefined reference to `__log10_finite'``. I wasn't able to reproduce this with the llvm/clang-9 toolchain on my machine. This seems to be a new error though, since I was previously able to emerge the llvm/clang-10 toolchain without errors. Perhaps this is due to a newer gcc/glibc?


InBetweenNames:
Thank you!

PR #516, opened 2020-05-09
Some Ofast and mtls-dialect=gnu2 workarounds

neeshy:
5 workarounds, included are:

media-video/mpv: disable ffast-math, a black screen was occurring when using the mitchell downscaler, or when using interpolation
sys-libs/glibc: disable ffast-math, same reason as opus (the preprocessor throws an error if ffast-math is enabled)
media-sound/mumble: disable ffast-math, same reason as opus (libopus is included as a bundled dependency)
net-libs/libtorrent-rasterbar: disable mtls-dialect=gnu2, this was causing memory corruption
www-client/torbrowser: disable mtls-dialect=gnu2, this one makes torbrowser and firefox workarounds consistent


InBetweenNames:
Thank you!

PR #517, opened 2020-05-09
Testworkarounds

Jannik2099:
These tests were done on a x86_64 glibc system.
Note that app-emulation/libvirt failed with those changes regardless, due to sandbox issues at a later point. All others passed tests with these changes


InBetweenNames:
Thank you!!

PR #522, opened 2020-05-13
net-misc/dhcp fix fatal runtime failure

Jannik2099:
dhclient (at least when dispatched by networkmanager) would send broken DHCPDISCOVERs , rendering it useless


InBetweenNames:
Thank you!

PR #529, opened 2020-05-24
My list of libtool overrides

TheGreatMcPain:
I upgraded to GCC 10 and did an `emerge -ave @world` this week.

I also have a bunch of `*FLAGS+="-fcommon"` entries, but I've excluded them, because I don't know where to put them yet.


InBetweenNames:
I tested this out and it looks like most of these actually work on my system, except for `media-gfx/imagemagick`.  There's a couple I can't test for reasons that are not related to `-fno-common`.  Could you post a `build.log` of `dev-lang/orc` failing?

TheGreatMcPain:
Oodly enough dev-lang/orc didn't fail for me, so I've removed it from the list.

InBetweenNames:
Could you supply a `build.log` from any of the other packages, say `dev-util/pkgconf`?

TheGreatMcPain:
I'm currently in the process of updating my list and I'm not sure how, but most of them are working again without the override.  Might be because I've recently upgrade glibc and binutils to their latest ~amd64 versions.

I'll post some logs if I run into a package that actually fails.

TheGreatMcPain:
I've re-emerged all the packages that have `NOCOMMON_LIBTOOL_OVERRIDE=1`, and all of them compiled without the override.

TheGreatMcPain:
I downgraded `binutils` back to 2.33.1-r1, and the errors started to show up again.
I guess `NOCOMMON_LIBTOOL_OVERRIDE=1` is not needed for `binutils-2.34`.

Here's a build.log of `dev-lang/orc` failing: [build.log.gz](https://github.com/InBetweenNames/gentooLTO/files/4739205/build.log.gz)

InBetweenNames:
Interesting -- I have had `sys-devel/binutils-2.34` since May 20th, but still need that workaround for a few packages.

Titaniumtown:
This fixed an issue with pkgconf compiling, thanks!!

Titaniumtown:
Will this be merged soon? I see nothing blocking it.

TheGreatMcPain:
I'll probably update this list after re-testing it now that binutils-2.34 is stable.

PR #549, opened 2020-07-16
readme / make.conf.lto touchups

### LINKED ISSUE #421: Lacking required gcc USE flags in README
Correct me, if this is just a misconfiguration of my own system, but it seems that applying the `make.conf` changes described in the README without having your active gcc compiled with `lto pgo graphite` (or at least one of these - have not checked combinatorially) results in `C compiler cannot create executables` on any non-gcc emerge attempt. These flag(s) are not mentioned in the README as required.  

Does this requirement make sense and maybe can comebody reproduce on their own system?

telans:
- <del>Removes translated readme files as they're outdated - and I don't think they're used. (GitHub insights view traffic?)</del>
 - Reorders make.conf.lto a little
 - Mentions the graphite use flag in the readme

Closes #471 
Closes #421


telans:
Updated

InBetweenNames:
The translated READMEs were contributed by a member here and it seems like they update them periodically.  I'm OK with moving them to a separate directory if people feel they are in the way -- I don't want to get rid of them entirely though.  Otherwise these changes look good to me.

telans:
@InBetweenNames Fair enough, how about this?

I've moved the translated files & templates to `.github` as well as fixing the links between them. Other config/readme changes are the same

InBetweenNames:
I like it!  Thank you!

PR #551, opened 2020-07-17
disable LTO on dev-scheme/gambit

vkraven:
Gambit is a dialect of scheme that compiles scheme code to C, and calls GCC to compile it to object files or binaries. 

The error occurs at runtime. When `gsc` is built with `-flto=*` or `-ffat-lto-objects` on GCC version 9 and above, `gsc` appears to pass object files to `ld` without escaping, which causes shell expansion and build failures. E.g.
```
... compile /home/user/git/gerbil/bootstrap/lib/gerbil/core$_MOP_$_MOP_1___rt.scm
x86_64-pc-linux-gnu-gcc: error: /tmp/core/usr/bin/x86_64-pc-linux-gnu-gccMOP_/usr/bin/x86_64-pc-linux-gnu-gccMOP_1___rt.o1.xaubLa.ltrans0.o: No such file or directory
x86_64-pc-linux-gnu-gcc: warning: ‘-x lto’ after last input file has no effect
x86_64-pc-linux-gnu-gcc: fatal error: no input files
```

Where `$_` gets expanded to `/usr/bin/x86_64-pc-linux-gnu-gcc`

`gsc` has no trouble with `-O3`, `-fipa-pta`, `-fno-semantic-interposition` or other optimization flags (it seems). 

PR to override `-flto` on `dev-scheme/gambit`. Thanks for reviewing!

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
+++ b/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf
@@ -96,7 +96,6 @@ sys-fabric/libibverbs *FLAGS-=-flto* # Issue #506, Undefined references
 app-office/gnucash *FLAGS-=-flto* # error: type ‘struct _iterate’ violates the C++ One Definition Rule [-Werror=odr]
 cross-i686-w64-mingw32/mingw64-runtime *FLAGS-=-flto* # linking errors
 cross-x86_64-w64-mingw32/mingw64-runtime *FLAGS-=-flto* # linking errors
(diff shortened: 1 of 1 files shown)


InBetweenNames:
Wow, thank you digging in and finding the real problem!  I would be happy to accept this PR -- just one minor change if that is OK.  Could you move this to the "LTO not recommended" section instead of the build failures section?  Even though it is technically a build failure, it's at package build rather than at runtime and I just want to keep it consistent.  Thank you!

vkraven:
Absolutely! It definitely makes more sense there. Modifications made, thanks for reviewing!

InBetweenNames:
Thank you!

Issue #553, opened 2020-07-20
sys-config/ltoize USE=keep-nocommon dosym for nocommon.conf missing line continuation


corrmaan:
```
 * Messages for package sys-config/ltoize-0.9.6:

 * Package:    sys-config/ltoize-0.9.6
 * Repository: lto-overlay
 * Maintainer: lookatyouhacker@gmail.com
 * USE:        abi_x86_64 amd64 elibc_glibc keep-nocommon kernel_linux userland_GNU
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
 * FEATURES='assume-digests binpkg-docompress binpkg-dostrip binpkg-logs config-protect-if-modified distlocks ebuild-locks fixlafiles ipc-sandbox merge-sync multilib-strict network-sandbox news parallel-fetch pid-sandbox preserve-libs protect-owned qa-unresolved-soname-deps sandbox sfperms split-elog strict unknown-features-warn unmerge-logs unmerge-orphans userfetch userpriv usersandbox usersync xattr'
 * CFLAGS='-march=skylake -mabm -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=16 -fuse-linker-plugin -pipe -falign-functions=32 -Wl,-O1 -Wl,--as-needed'
 * CXXFLAGS='-march=skylake -mabm -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=16 -fuse-linker-plugin -pipe -falign-functions=32 -Wl,-O1 -Wl,--as-needed'
 * FFLAGS='-march=skylake -mabm -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=16 -fuse-linker-plugin -pipe -falign-functions=32 -Wl,-O1 -Wl,--as-needed'
 * FCFLAGS='-march=skylake -mabm -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=16 -fuse-linker-plugin -pipe -falign-functions=32 -Wl,-O1 -Wl,--as-needed'
[...]

PR #554, opened 2020-07-20
sys-config/ltoize USE=keep-nocommon dosym for nocommon.conf fix

corrmaan:
Title: sys-config/ltoize: USE=keep-nocommon dosym for nocommon.conf fix

fixes https://github.com/InBetweenNames/gentooLTO/issues/553


InBetweenNames:
Thank you!

PR #555, opened 2020-07-20
allow lto for >=openjdk-14

Hello71:
for reasons beyond me, java 12+ is not in upstream gentoo, but I tested lto with 14.0.2 and it works fine, where 11.0.8 doesn't. I think this was fixed upstream, but I'm not completely sure how.


InBetweenNames:
Is this via custom ebuild or an overlay somewhere?

Hello71:
just cp openjdk-11.0.8_p10.ebuild openjdk-14.0.2_p12.ebuild. the make 4.3 patch was required in 14.0.1 but not in 14.0.2.

InBetweenNames:
Seems to be working here!  Thanks!

PR #561, opened 2020-07-30
app-portage/lto-rebuild: add arm64 keyword

dblaber:
lto-rebuild works fine on arm64, so adding ~arm64 to avoid ~* and ** in
arm64 users keyword file.

Title: app-portage/lto-rebuild: add arm64 keyword

lto-rebuild works fine on arm64, so adding ~arm64 to avoid ~* and ** in


InBetweenNames:
Wonderful!  Glad to see more ARM users here 👍

PR #564, opened 2020-07-31
Suggest NTHREADS="auto" in documentation and samples

### LINKED ISSUE #527: Automatically assigning NTHREADS ?
I use the same /etc/portage on all of my machines (Stored in a git repository).

I want to have NTHREADS calculated based on how many cores are on a machine, but can't with the current method of things, as far as I'm able to tell.

Is this possible?

### LINKED ISSUE #565: LLVM Merges and many others use over NTHREADS for LTO cycle
GCC 10 and higher support -flto=auto, allowing for gcc to determine optimal number of cores to use. If -flto is set to a number, and certain projects such as llvm are build, it will multiply this number by number of make instances, causing too many processes to be created, ultimately slowing down the build to a crawl on smaller devices with less ram(or causing OOM conditions).  We should update the sample config  and readme to suggest the "auto" option. 

While emerging llvm 10.0.1
NTHREADS set to 4 in /etc/portage/make.conf
MAKEOPTS="-j21 -l4"

dmb@pi4 ~ $ ps aux|grep lto1|wc -l
12

It is observed mutliple make instances are running in this case, each with -flto=4, so 4*number of make instances can be too many.

dblaber:
GCC 10 and higher support -flto=auto, allowing for gcc to determine
optimal number of cores to use. If -flto is set to a number, and certain
projects such as llvm are build, it will multiply this number by number
of make instances, causing too many processes to be created, ultimately
slowing down the build to a crawl on smaller devices with less ram(or
causing OOM conditions). This commit suggests using auto in configs for
ltoize and README.md.

Title: sys-config/ltoize: Suggest NTHREADS="auto" in documentation and samples
Closes #565 
Closes #527

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ The biggest gotcha with `-O3` is that it does not play nice at all with Undefine
 Add the `mv` overlay (`layman -a mv`) and then add this overlay (`layman -a lto-overlay`) to your system and run `emerge sys-config/ltoize`. Add the package to your `/etc/portage/package.accept_keywords` if necessary. This will add the necessary overrides to `/etc/portage/`, but it won't modify your `make.conf`. It will create a `make.conf.lto` symlink in `/etc/portage` with the default GentooLTO configuration. To use the default configuration, define a variable `NTHREADS` with the number of threads you want to use for LTO. Then, source the file in your own `make.conf` like in this example:
 
 ~~~ bash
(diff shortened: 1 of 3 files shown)


InBetweenNames:
I've been using this myself for a little while -- thank you!  I forgot to update these parts of the docs myself back when we did the GCC 10 update.

PR #567, opened 2020-08-06
Re-enable LTO for Mesa

Althorion:
As per https://github.com/InBetweenNames/gentooLTO/issues/143 Mesa could now be safely compiled with LTO and that’s in fact a default for some other distros, like Arch.


InBetweenNames:
Thanks!  We may need to revert this but I want to try it out.

barolo:
Can't affected ones be excluded by use flag? I remember seeing it discussed somewhere here. Similar situation with qemu
@InBetweenNames

PR #569, opened 2020-08-11
ltoworkarounds: remove a fdevirtualize-at-ltrans workaround

telans:
Cannot reproduce ICE


InBetweenNames:
Looks like it was fixed in GCC 10!  Thanks for the PR!

PR #571, opened 2020-08-12
Fix repoman warnings, remove dev-python/notify-python

telans:
- updated ltoize copyrights
- removed portage-mgorny dependency (missing from tree/overlays?)
- reorder KEYWORDS
- dev-python/notify-python was removed as it just needs `NOCOMMON_OVERRIDE_LIBTOOL`


InBetweenNames:
Thank you!

PR #573, opened 2020-08-12
sys-config/ltoize: Made the warn message more understandable

Kreyren:
**Title:** sys-config/ltoize: Made the warn message more understandable

Community member (@uniminin) asked me for help with with this ebuild that was failing because he had version 9.3.0 installed where i believe that the current method could be confusing for some users thus reasoning for this merge request.

Signed-off-by: Jacob Hrbek <kreyren@rixotstudio.cz>

### DIFF
--- a/sys-config/ltoize/ltoize-0.9.2.ebuild
+++ b/sys-config/ltoize/ltoize-0.9.2.ebuild
@@ -37,9 +37,9 @@ pkg_setup() {
 	ACTIVE_GCC=$(gcc-fullversion)
 
 	if ver_test "${ACTIVE_GCC}" -lt 8.2.0; then
-		ewarn "Warning: Active GCC version < 8.2.0, it is recommended that you use the newest GCC if you want LTO."
+		ewarn "Warning: Active GCC version '$ACTIVE_GCC' is lower then the expected version '8.2.0', it is recommended that you use the newest GCC if you want LTO."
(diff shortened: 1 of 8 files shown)


Kreyren:
i've overwrote the logic by mistake, fixing

sys-config/ltoize/ltoize-0.9.2.ebuild
```diff
@@ -37,9 +37,9 @@ pkg_setup() {
 	ACTIVE_GCC=$(gcc-fullversion)
 
 	if ver_test "${ACTIVE_GCC}" -lt 8.2.0; then
-		ewarn "Warning: Active GCC version < 8.2.0, it is recommended that you use the newest GCC if you want LTO."
+		ewarn "Warning: Active GCC version '$ACTIVE_GCC' is lower then the expected version '8.2.00', it is recommended that you use the newest GCC if you want LTO."
```
telans, line 40:
```suggestion
		ewarn "Warning: Active GCC version '$ACTIVE_GCC' is lower then the expected version '8.2.0', it is recommended that you use the newest GCC if you want LTO."
```

Kreyren: @telans good catch! Resolved

InBetweenNames:
I like this!

InBetweenNames:
Thank you!

PR #576, opened 2020-08-16
workaround: net-libs/libiscsi fails to compile with -fno-semantic-interposition

glitchking007:
error: inlining failed in call to ‘always_inline’ ‘read.localalias’: function not inlinable

Compiles with workaround


InBetweenNames:
Thank you!

Issue #600, opened 2020-09-11
app-arch/zstd fails to build on my system


gorrila20:
app-arch/zstd fails to build (in prepare phase) when doing a world update. 
Maybe I just did something stupid, but some help will be appreciated anyways!
::

>>> Unpacking source...
>>> Unpacking zstd-1.4.4.tar.gz to /var/tmp/portage/app-arch/zstd-1.4.4-r4/work
>>> Source unpacked in /var/tmp/portage/app-arch/zstd-1.4.4-r4/work
 * Applying zstd-1.4.5-multi-thread-default.patch ...
patching file programs/zstdcli.c
Hunk #1 succeeded at 624 with fuzz 1 (offset -71 lines).
patching file programs/README.md
Hunk #1 FAILED at 168.
1 out of 1 hunk FAILED -- saving rejects to file programs/README.md.rej
patching file programs/zstdcli.c
Hunk #1 FAILED at 160.
1 out of 1 hunk FAILED -- saving rejects to file programs/zstdcli.c.rej                             [ !! ]
 * ERROR: app-arch/zstd-1.4.4-r4::gentoo failed (prepare phase):
 *   patch -p1  failed with /var/lib/layman/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch
 *
 * Call stack:
 *               ebuild.sh, line  762:  Called __ebuild_main 'prepare'
 *      phase-functions.sh, line  970:  Called __source_all_bashrcs
 *               ebuild.sh, line  409:  Called __try_source '--no-qa' '/etc/portage/bashrc'
 *               ebuild.sh, line  458:  Called source '/etc/portage/bashrc'
 *                  bashrc, line    6:  Called BashrcdMain '/etc/portage/bashrc'
 *             environment, line  598:  Called LTOPatch
 *             environment, line 1265:  Called eapply '/var/lib/layman/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch'
 *             environment, line 1812:  Called _eapply_patch '/var/lib/layman/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch'
[...]

PR #601, opened 2020-09-12
app-arch/zstd: Create version specific patches for multithreaded compression

jiblime:
app-arch/zstd: Create version specific patches

1.4.4-r4 fails to patch with 1.4.5 patch

This makes the multi-thread-default.patch know to apply the correct patch depending on version. Only the latest stable (1.4.4-r4) and latest unstable (1.4.5) have patches created.

Fixes: https://github.com/InBetweenNames/gentooLTO/issues/600


jiblime:
Sidenote:

How do I merge commits together cleanly? I thought force pushing would erase previous history, didn't mean to create multiple commit references...

image

jiblime:
Closing on the basis that lto-overlay users should be keywording `~amd64` by default. I still would like help with the aforementioned, however...

PR #603, opened 2020-09-13
github: use workflows, remove travis

telans:
Runs on every push (including pull requests).

See an example here: https://github.com/telans/gentooLTO/runs/1107511533?check_suite_focus=true


Titaniumtown (approved):
LGTM. Good idea!

InBetweenNames:
Oh I quite like this!  Thank you!  Lets try this out and see how it goes.

PR #611, opened 2020-10-07
dev-libs/rocr-runtime: Build without LTO

2l42h3r:
Building dev-libs/rocr-runtime will cause tools like rocminfo or clinfo (and presumably many others) to segfault.
Disabling LTO completely fixes the issue (when built with fat lto objects the issue still persists).


2l42h3r:
Just realized I put the fix in the wrong category. Will reopen another pr.

PR #612, opened 2020-10-07
dev-libs/rocr-runtime: Build without LTO

2l42h3r:
Building rocr-runtime with LTO causes tools like rocminfo or clinfo (and presumably many others) to segfault at runtime.
Had to disable LTO completely, building with fat objects didn't solve the issue.


InBetweenNames:
Thank you!

PR #615, opened 2020-10-08
dev-python/libvirt-python Disable LTO on versions >=6.8.0 due to ICE

petronio:


TheGreatMcPain:
I can confirm that this happens, but it was on version version 6.7.0

Although, It did successfully merge on the second attempt without this workaround.

InBetweenNames:
Thank you!

PR #618, opened 2020-10-12
Fix segfault when starting libvirtd

aruhier:
Add libvirt in the -fipa-pta workarounds to fix a segfault when starting libvirt daemon.

Without this fix:
```
$ gdb libvirtd                                      
(gdb) run
Starting program: /usr/sbin/libvirtd
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7c06530 in virLogOutputListFree () from /usr/lib64/libvirt.so.0
```

With this fix: libvirtd starts normally.


aruhier:
FYI: this bug started to appear when I rebuilt libvirt for the 6.8.0 update, but I have the now same issue by rolling back to 6.7.0. I didn't have this issue before on 6.7.0 (before the 3rd of October), so I don't know which change produced this issue.

InBetweenNames:
Thanks -- this is a pretty serious one.  We'll disable it on all `libvirt` packages until more investigation is done.

PR #621, opened 2020-10-18
net-misc/nextcloud-client: Fixes segfault by turning off LTO

aruhier:
Desactivate LTO for nextcloud-client to fix a segfault at start.

Without this fix:
```
Thread 1 "nextcloud" received signal SIGSEGV, Segmentation fault.
0x00007fffeec6f861 in ?? () from /usr/lib64/libQt5Core.so.5
```

With this fix: `nextcloud` starts normally.


Titaniumtown (approved):
LGTM

InBetweenNames:
Thank you!

Issue #602, opened 2020-09-12
media-libs/x264 fails to build during prepare phase


gorrila20:
media-libs/x264 fails to build on my system while performing an @world update.

Failed attempts I took to fix the issue:

- Emerging ~amd64 package
- Emerging 9999 package

The lto patch fails in every package version, except for the 9999 one where it gives me a little endian check error.

Error log of normal package no ~amd64:

 * x264-snapshot-20190214-2245.tar.bz2 BLAKE2B SHA512 size ;-) ...                                                          [ ok ]
 * FEATURES='assume-digests binpkg-docompress binpkg-dostrip binpkg-logs config-protect-if-modified distlocks ebuild-locks fixlafiles ipc-sandbox merge-sync multilib-strict network-sandbox news parallel-fetch pid-sandbox preserve-libs protect-owned qa-unresolved-soname-deps sandbox sfperms strict unknown-features-warn unmerge-logs unmerge-orphans userfetch userpriv usersandbox usersync xattr'
 * CFLAGS='-march=native -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=12 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed'
 * CXXFLAGS='-march=native -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=12 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed'
 * FFLAGS='-march=native -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=12 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed'
 * FCFLAGS='-march=native -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=12 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed'
 * F77FLAGS='-march=native -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=12 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed'
[...]

PR #623, opened 2020-10-19
media-libs/x264: Minor fix for issue #602

powpingdone:
No patches needed. You can just use the least amount of effort by making it fat.

Provides workaround for #602


Titaniumtown:
#597? This has already been resolved I think.

PR #625, opened 2020-10-21
dev-python=>3.7.9: use --pgo-extended

### LINKED ISSUE #552: use built-in python lto/pgo?
I used these options with standard Gentoo ebuild:

```
export PROFILE_TASK="-m test -j $(nproc) --pgo-extended"
EXTRA_ECONF="--with-lto --enable-optimizations"
```

telans:
Closes #552 

I haven't actually tested them all but the diffs are the same


InBetweenNames:
Thank you!  This seems reasonable -- but we'll have to see if anyone complains.

PR #628, opened 2020-10-22
dev-util/sccache: Fails to link without -ffat-lto-objects

TheGreatMcPain:
Similar to other rust programs, `sccache` needs `-ffat-lto-objects` to build.


InBetweenNames:
Thank you!

Issue #622, opened 2020-10-18
app-admin/keepassxc segfaults when starting since Qt 5.15.1


kotelnik:
Adding `app-admin/keepassxc *FLAGS-=-flto*` to `ltoworkarounds.conf` resolves the issue.

Titaniumtown:
@kotelnik you should probably make a PR then....

aruhier:
Weird, I don't have the issue.

Which flags are you using?

Here is what I'm using:
```
+ autotype
+ browser 
- ccache  
- debug   
- keeshare
+ network 
- test    
- yubikey
```

telans:
> Here is what I'm using:

KeePassXC isn't built with lto if you have the `autotype` USE flag set

https://github.com/InBetweenNames/gentooLTO/blob/46962ee83cd32b186b5a84f8963586c0a8346aeb/sys-config/ltoize/files/package.cflags/ltoworkarounds.conf#L33

aruhier:
Oh, makes sense. Thanks!

PR #629, opened 2020-10-23
app-admin/keepassxc: disable lto, otherwise segfaults on start

kotelnik:
Disable lto entirely, otherwise it segfaults on start since Qt 5.15.1

Fixes issue:
https://github.com/InBetweenNames/gentooLTO/issues/622


InBetweenNames:
Confirmed -- thank you!

PR #633, opened 2020-10-30
Remove dependency on `mv` overlay

### LINKED ISSUE #610: Suggestion: don't require `mv` overlay
The `mv` overlay is currently being used to provide a way to extend portage through `/etc/portage/bashrc.d`. This works pretty nicely for the purposes of this project, as the flag overrides file is really nice. However, I don't use this overlay, or the other features of this bashrc.d implementation, and I'd like to keep my portage's behavior as "default" as possible. Having to package.mask the overlay is a bit of a nuisance, and there's no obvious way to disable certain functionality of this bashrc.d.

That leads me to create this suggestion. It'd be nice if the ltoize ebuild could (optionally) include a stripped-down version of portage-bashrc-mv. That would ideally just include `bashrcd.sh` and `10-flag.sh`, and have a USE flag to instead use the "full" version by depending on the ebuild in the `mv` overlay. This would avoid having to add an overlay full of unrelated ebuilds and unexpected portage behavior related to this bashrc.d.

telans:
`app-portage/portage-bashrc-mv` has been modified from the `mv` overlay, installing only files provided by the `cflags` USE flag (plus required files).

I've updated `sys-config/ltoize` a little, removing the verbose logging and the binutils check. I feel the `elog` lines provided no required information when it's all explained in `pkg_postinst` and the readme anyway. The binutils check was removed as 2.34 has been stabilized.
@InBetweenNames I put you as the maintainer within `metadata.xml` for `portage-bashrc-mv`, if that isn't an issue.

Closes #610 

---

Repoman is passing, but coming up with:
```
    ebuild.badheader              1
   app-portage/portage-bashrc-mv/portage-bashrc-mv-20.2.ebuild: line 1: Invalid Copyright
```
This is due to the addition of  `and others` within the copyright line of the ebuild. This shouldn't be a "bad header", as mgorny seems to think it's alright in https://mgorny.pl/articles/new-gentoo-copyright-policy-explained.html#copyright-ownership


epsilon-0:
This would be nice to have :heart:

telans:
Regardless of whether @InBetweenNames likes the changes I've made to the ebuilds, I definitely think it's bad practice to rely on packages from a third party overlay we don't control (sorry vaeth)

ignamartinoli:
Agree with @telans

PR #634, opened 2020-10-31
lto-rebuild: fix rdepends, install phase

### LINKED ISSUE #620: lto-rebuild has incorrect install phase
lto-rebuild currently does this:

pkg_preinst() {
    dobin "${FILESDIR}"/lto-rebuild
}

however, ${FILESDIR} is not guaranteed to be populated in any non-src phase. This is an issue when installing the package from a PORTAGE_BINHOST

the solution is to just do

src_install() {
    dobin "${FILESDIR}"/lto-rebuild
}

Additionally, lto-rebuild has some dependencies that are in @system :

app-shells/bash:*
sys-devel/binutils:*
sys-devel/gcc:*

Packages are not supposed to depend on @system packages except for USE toggles, conditional blockers or others.
See https://devmanual.gentoo.org/general-concepts/dependencies/#implicit-system-dependency

telans:
Closes #620


InBetweenNames:
Thank you!

PR #635, opened 2020-11-02
Update patch for media-libs/x264-0.0.20190903-r1

iphands:
Title: media-libs/x264: lto.patch fails to apply to >= 20190903-r1

Simple fixing patch that applies -shared to CC when doing an endianess check

### DIFF
--- a/sys-config/ltoize/files/patches/media-libs/x264/lto.patch
+++ b/sys-config/ltoize/files/patches/media-libs/x264/lto.patch
@@ -1,11 +1,12 @@
---- a/configure 2020-07-07 20:37:00.103200843 -0400
-+++ b/configure 2020-07-07 20:36:38.819867863 -0400
-@@ -1017,7 +1017,7 @@
+diff -uw a/configure b/configure
+--- a/configure	2020-11-02 12:24:41.569046153 -0500
++++ b/configure	2020-11-02 12:24:34.609046389 -0500
+@@ -994,7 +994,7 @@
  CPU_ENDIAN="little-endian"
  if [ $compiler = GNU ]; then
      echo "int i[2] = {0x42494745,0}; double f[2] = {0x1.0656e6469616ep+102,0};" > conftest.c
 -    $CC $CFLAGS conftest.c -c -o conftest.o 2>/dev/null || die "endian test failed"
 +    $CC $CFLAGS conftest.c -c -o conftest.o -shared 2>/dev/null || die "endian test failed"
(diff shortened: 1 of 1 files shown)


iphands:
This also works for 0.0.20190214

telans:
You should mirror this to the x264-encoder patch as well

iphands:
@telans hey I was playing around with this some...
I ended up realizing I goofed a bit and put the patch in for x264-encoder in the first place.

In addition to that though... while testing I dont thing adding `-shared` to the conftest fixes this for me on my box.
Im curious... the endianness check... is there any real reason to use $CFLAGS there?

I am (naively) wondering if this is still as good as checking local endianess but gets tripped up by cflags less:
```diff
-    $CC $CFLAGS conftest.c -c -o conftest.o 2>/dev/null || die "endian test failed"
+    $CC -O0 conftest.c -c -o conftest.o -shared 2>/dev/null || die "endian test failed ONE"
```

Wondering if there are any gotchas there at all?

iphands:
FWIW this is what works for me:
https://github.com/InBetweenNames/gentooLTO/compare/master...iphands:tmp_O0_patch

Can push up here, but lemme know if this is not okay.

telans:
Sorry for the delay. I would change the patch to:

```
--- a/configure	2020-11-11 12:42:21.977274237 -0500
+++ b/configure	2020-11-11 12:55:30.545286800 -0500
@@ -994,7 +994,7 @@
 CPU_ENDIAN="little-endian"
 if [ $compiler = GNU ]; then
     echo "int i[2] = {0x42494745,0}; double f[2] = {0x1.0656e6469616ep+102,0};" > conftest.c
-    $CC $CFLAGS conftest.c -c -o conftest.o 2>/dev/null || die "endian test failed"
+    $CC -O0 conftest.c -c -o conftest.o 2>/dev/null || die "endian test failed"
     if (${STRINGS} -a conftest.o | grep -q BIGE) && (${STRINGS} -a conftest.o | grep -q FPendian) ; then
         define WORDS_BIGENDIAN
         CPU_ENDIAN="big-endian"
```

This requires the gentoo patches to be applied first, so this PR would depend on #653 . I believe this is a better way of applying patches (on top of whatever the gentoo tree does), but if you just want this merged without waiting for that PR the patch as you have it is good:

> FWIW this is what works for me:
> [master...iphands:tmp_O0_patch](https://github.com/InBetweenNames/gentooLTO/compare/master...iphands:tmp_O0_patch)

InBetweenNames:
Fixed via 911a0a1850a59f6e5594636404a51762188356b1

PR #644, opened 2020-11-11
Disable LTO for net-news/newsboat

Titaniumtown:
Title: net-news/newsboat: disable LTO

Disable LTO for net-news/newsboat, package fails to build with LTO.


InBetweenNames:
Thank you!

PR #646, opened 2020-11-12
Disable LTO for app-text/yodl

Titaniumtown:
Title: app-text/yodl:  Disable LTO

yodl doesn't build with LTO, this PR disables LTO so yodl can be compiled.


InBetweenNames:
Thank you!

PR #649, opened 2020-11-16
separate workarounds into specific files per flag

telans:
Makes searching through workarounds simpler. File names seem alright to me


InBetweenNames:
I agree this is the way to go.  Thanks for taking the time to do it -- I'll need to review this a bit more thoroughly over the coming days but I expect to merge it soon.

telans:
rebased

PR #664, opened 2020-12-17
Issue #663

jbcjorge:
This closes #663, as the patch doesn't work for the 1.4.7 version, a
further patch will be submited to address that

Issue #663, opened 2020-12-17
LTOPatch fails to apply on app-arch/zstd-1.4.7


thrifty-txt:
``` * ERROR: app-arch/zstd-1.4.7::gentoo failed (configure phase):
 *   patch -p1  failed with /var/db/repos/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch
 *
 * Call stack:
 *               ebuild.sh, line  762:  Called __ebuild_main 'configure'
 *      phase-functions.sh, line  970:  Called __source_all_bashrcs
 *               ebuild.sh, line  409:  Called __try_source '--no-qa' '/etc/portage/bashrc'
 *               ebuild.sh, line  458:  Called source '/etc/portage/bashrc'
 *                  bashrc, line    6:  Called BashrcdMain '/etc/portage/bashrc'
 *             environment, line  608:  Called LTOPatch
 *             environment, line 1275:  Called eapply '/var/db/repos/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch'
 *             environment, line 1851:  Called _eapply_patch '/var/db/repos/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch'
 *             environment, line 1789:  Called __helpers_die 'patch -p1  failed with /var/db/repos/lto-overlay/sys-config/ltoize/files/patches/app-arch/zstd/zstd-1.4.5-multi-thread-default.patch'
 *   isolated-functions.sh, line  112:  Called die
```

InBetweenNames:
Should be working now -- just comment again if the issue persists and I'll reopen.

PR #665, opened 2020-12-17
Enable multiprocessor for zstd by default

jbcjorge:
This applies to app-arch/zstd-1.4.7, enabling the multiprocessor as default option

It's the continuation of #664 and #663

### DIFF
--- a/sys-config/ltoize/files/patches/app-arch/zstd-1.4.7/zstd-1.4.7-multi-thread-default.patch
+++ b/sys-config/ltoize/files/patches/app-arch/zstd-1.4.7/zstd-1.4.7-multi-thread-default.patch
@@ -1,0 +1,23 @@
+--- a/programs/README.md
++++ b/programs/README.md
+@@ -172,7 +172,7 @@
+ --long[=#]: enable long distance matching with given window log (default: 27)
+ --fast[=#]: switch to very fast compression levels (default: 1)
+ --adapt : dynamically adapt compression level to I/O conditions
+- -T#    : spawns # compression threads (default: 1, 0==# cores)
++ -T#    : spawns # compression threads (default: 0==# cores)
+  -B#    : select size of each job (default: 0==automatic)
+ --single-thread : use a single thread for both I/O and compression (result slightly different than -T1)
+ --rsyncable : compress using a rsync-friendly method (-B sets block size)
+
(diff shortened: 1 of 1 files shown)


Jannik2099:
1. does zstd pass its unit tests with this?
2. does this affect CLI behaviour, as in are there cases where it breaks scripts?
3. Couldn't we define -DZSTDCLI_NBTHREADS_DEFAULT=1 in package.cflags instead of patching?

jiblime:
@Jannik2099 
1. With or without this patch, I fail to pass tests for 1.4.7, even when specifying `*FLAGS="-march=native -O2 -pipe"`. I was able to pass all tests on 1.4.5. IIRC I needed to remove `${GRAPHITE}` flags.

2. It's rather simple so it shouldn't, and I haven't encountered any issues. If you want it to be single-threaded, you may manually specify so; I'm assuming by breaking scripts you are referring to build configs that override for single-threaded compression. 

3. That may be a better solution. The patch would still be needed for clarity since calling `--help` would give incorrect information. 

[zstd-1.4.7_fail_gcc1021.log](https://github.com/InBetweenNames/gentooLTO/files/5735857/zstd-1.4.7_fail_gcc1021.log)


Unrelated: there's something wrong with the upstream build (or possibly my machine)
```
 167   │ make -j12 CC=x86_64-pc-linux-gnu-gcc CXX=x86_64-pc-linux-gnu-g++ AR=x86_64-pc-linux-gnu-ar PREFIX=/usr LIBDIR=/usr/lib64 -C lib libzstd-mt libzstd.a-mt libzstd.pc 
 168   │ make: Entering directory '/var/tmp/portage/app-arch/zstd-1.4.7/work/zstd-1.4.7-abi_x86_64.amd64/lib'
 169   │ make[1]: Entering directory '/var/tmp/portage/app-arch/zstd-1.4.7/work/zstd-1.4.7-abi_x86_64.amd64/lib'
 170   │ make[1]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
```

InBetweenNames:
Thank you!  Just going to move it over to 1.4.8

InBetweenNames:
I agree adding the preprocessor define directly would be simpler, but I do like that `--help` gives the right information with this patch.

Issue #671, opened 2020-12-20
media-libs/x264-9999: endian test failed


nvertigo:
In this commit "-c" is (re-)intoduced to the conftest compilation. This breaks the endian test for my environment. Dropping "-c" again fixes the issue:

[x264-fix-endian-test.zip](https://github.com/InBetweenNames/gentooLTO/files/5720935/x264-fix-endian-test.zip)
or
https://github.com/InBetweenNames/gentooLTO/pull/672

"-c" jas been introduced here: https://github.com/InBetweenNames/gentooLTO/commit/e96d39e31b5f2b2996c570f39c4638cc6c0a45da

The same for x264-encoder - patch updated.

nvertigo:
Just a quick test:

adler tmp # echo "int i[2] = {0x42494745,0}; double f[2] = {0x1.0656e6469616ep+102,0};" > conftest.c
adler tmp # x86_64-pc-linux-gnu-gcc conftest.c -march=bdver1 -mtune=bdver1 -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=8 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu -Wall -I. -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu -march=bdver1 -c -shared -o conftest-with_-c.o
adler tmp # x86_64-pc-linux-gnu-gcc conftest.c -march=bdver1 -mtune=bdver1 -O3 -fgraphite-identity -floop-nest-optimize -fdevirtualize-at-ltrans -fipa-pta -fno-semantic-interposition -flto=8 -fuse-linker-plugin -pipe -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu -Wall -I. -Wl,-O1 -Wl,--as-needed -Wl,--hash-style=gnu -march=bdver1 -shared -o conftest-without_-c.o
adler tmp # strings conftest-with_-c.o |grep -e EGIB -e naidnePF -e BIGE -e FPendian
adler tmp # strings conftest-without_-c.o |grep -e EGIB -e naidnePF -e BIGE -e FPendian
EGIB
naidnePF
adler tmp #

jiblime:
[...]

PR #672, opened 2020-12-20
media-libs/x264|-encoder/lto.patch: fix failing endian test.

nvertigo:
Without linking ("-c") the endian tests fail for:
  media-libs/x264 and
  media-libs/x264-encoder

  See: https://github.com/InBetweenNames/gentooLTO/issues/671


InBetweenNames:
Merged!  Thank you!

PR #674, opened 2020-12-22
fix broken gcc due to gmp on arm64

Jannik2099:
This fixes gmp-6.2.1 on arm64 - without, it breaks with lto, resulting in e.g. gfortran being stuck in an infinite loop

Patch is submitted upstream aswell - I'll submit to ::gentoo once upstream explains what's going on and wether this is the right approach

### DIFF
--- a/sys-config/ltoize/files/patches/dev-libs/gmp/arm64-lto.patch
+++ b/sys-config/ltoize/files/patches/dev-libs/gmp/arm64-lto.patch
@@ -1,0 +1,45 @@
+diff -r 575ff753614f mpn/arm64/bdiv_q_1.asm
+--- a/mpn/arm64/bdiv_q_1.asm	Fri Jan 01 14:46:22 2021 +0100
++++ b/mpn/arm64/bdiv_q_1.asm	Fri Jan 01 17:13:01 2021 +0100
+@@ -61,9 +61,15 @@
+ 	clz	cnt, x6
+ 	lsr	d, d, cnt
+ 
+-	LEA_HI(	x7, binvert_limb_table)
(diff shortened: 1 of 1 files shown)


Jannik2099:
actually please hold up on merging this, currently looking at an issue

paralin:
@Jannik2099 funny enough just ran into this doing the pinebook compilation. :)

Jannik2099:
I haven't gotten an answer from upstream yet, so just revert the commit entirely - ready to merge now

InBetweenNames:
Did this end up getting merged upstream?  It's unclear if this is still intended to be merged or not here.

Jannik2099:
I reported the bug upstream - this just reverts the commit that caused this, I'm sure upstream will come up with a more elegant solution, but they haven't replied yet.

Until then please merge it, as it currently leads to a broken gcc :/

InBetweenNames:
Merged, thanks!  Can you link me to the upstream bug just so I can keep tabs on it?

Jannik2099:
It's a mailing list, original post https://gmplib.org/list-archives/gmp-bugs/2020-December/004981.html

Aransentin:
A more pernicious effect of this bug is that it breaks floating point literal parsing in GCC:

For example, this:
```#include <stdio.h>
int main (void)
{   
    printf( "%f\n", 10000.1 );
}
```

prints `9961.569252`. This causes miscompilations if you build other software with the broken gmp library.

paralin:
Good thing fortran gets stuck in an infinite loop so we know something is broken, rather than just using programs with inaccurate floats, heh.

Issue #524, opened 2020-05-14
games-fps/gzdoom-4.3.3::gentoo doesn't build with ninja or lto


nightloli:
i got 

ERROR: compile
ERROR: games-fps/gzdoom-4.3.3::gentoo failed (compile phase):
  ninja -v -j4 -l0 failed

Call stack:
    ebuild.sh, line  125:  Called src_compile
  environment, line 3559:  Called cmake_src_compile
  environment, line 1976:  Called cmake_build
  environment, line 1955:  Called eninja
  environment, line 2415:  Called die
The specific snippet of code:
      "$@" || die "${nonfatal_args[@]}" "${*} failed"

when i was trying to build games-fps/gzdoom with ninja
and i was getting segfault when i was trying to build it with either just -flto or -flto -ffat-lto-objects
It only compiled with

games-fps/gzdoom CMAKE_MAKEFILE_GENERATOR=emake
games-fps/gzdoom *FLAGS-=-flto*

elsandosgrande:
> It only compiled with
> 
> games-fps/gzdoom CMAKE_MAKEFILE_GENERATOR=emake
> games-fps/gzdoom FLAGS-=-flto

Do you need both, or is it just that either one will make the compilation succeed?

nightloli:
I needed both to successfully compile it

elsandosgrande:
In that case, can you compile GZDoom on a clean system, a Gentoo testing installation without ever having GentooLTO installed, without either environment variable?

InBetweenNames:
Indeed this seems to not work on GCC 9.  On GCC 10, I can't reproduce -- just built `games-fps/gzdoom` without any issues on that one.

TheGreatMcPain:
Using GCC 9 gzdoom builds just fine with ninja.  I'm also able to build it using LTO, but the resulting binary segfaults.

elsandosgrande:
@TheGreatMcPain Have you tried it with GCC 10? If so, what is your mileage?

TheGreatMcPain:
@elsandosgrande I haven't used it yet, but I plan to in the coming days.

InBetweenNames:
Is this the error you get?

~~~
GZDoom <unknown version> -  - SDL version
Compiled on May 15 2020

M_LoadDefaults: Load system defaults.
[...]

PR #682, opened 2021-01-08
games-fps/gzdoom: Disabled LTO + Force Emake

billyb2:
GZDoom doesn't build otherwise, unfortuanetly. According to [this issue](https://github.com/InBetweenNames/gentooLTO/issues/524), it also doesn't work with ninja. Although I didn't test this for myself, I disabled it too just in case.


InBetweenNames:
Thank you!

PR #685, opened 2021-01-12
games-fps/zandronum: Disabled LTO

billyb2:
Similar to GZDoom, Zandronum doesn't start when LTO is enabled. Though, interestingly, it will run, but it can't properly read Doom .wad files. Both GZDoom and Zandronum are based off of ZDoom, so that could be it.


InBetweenNames:
Thank you!  Yeah it looks like there's some UB in these source ports.

Issue #680, opened 2021-01-07
x264 and x264-encoder patch no longer needed for 9999 versions.


TheGreatMcPain:
This [commit](https://code.videolan.org/videolan/x264/-/commit/6bc7fe4f36ea95db77e6df6d76153dd5a2c770a0) was pushed to x264's repository today.

InBetweenNames:
Fixed via 530494cc1c3ef39ce75c7b68a26ec28ac75c2e37 - thanks!

PR #686, opened 2021-01-14
media-libs/x264|-encoder: Don't apply patch for live version

TheGreatMcPain:
The latest x264 master has added a workaround for LTO, so there's no
need to apply the LTO patch for newer versions of x264. See #680


InBetweenNames:
Wonderful!  Thank you!

PR #687, opened 2021-01-18
ltoworkarounds: remove media-libs/flac -fno-ipa-cp-clone

telans:
`flac -d` works fine without it on my system with gcc 10.2.0

It also breaks configure with clang, as clang doesn't support the flag.


InBetweenNames:
Thank you!

PR #691, opened 2021-01-22
dev-util/colm: disable remove-la from portage-bashrc-mv

### LINKED ISSUE #690: remove-la breaks colm and ragel
dev-util/colm-0.14.6 fails to build but works with noflto
but then
dev-util/ragel-7.0.3 fails to build needing libfsm.la

removing the remove-la flag both packs build fine

petronio:
Fixes #690

If you're having a dependency issue while attempting to rebuild it, first try downgrading colm and then updating ragel (pulls in newer colm).

### DIFF
--- a/sys-config/ltoize/files/package.cflags/portage-bashrc-mv.conf
+++ b/sys-config/ltoize/files/package.cflags/portage-bashrc-mv.conf
@@ -1,0 +1,3 @@
+# BEGIN: portage-bashrc-mv remove-la workarounds
+dev-util/colm NOLAFILEREMOVE=true # See issue 690 and https://bugs.gentoo.org/766210
+# END: portage-bashrc-mv remove-la workarounds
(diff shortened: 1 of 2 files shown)


telans:
Could you rebase and maybe make a new file such as `bashrc-mv.conf` for general overrides?

petronio:
> Could you rebase and maybe make a new file such as `bashrc-mv.conf` for general overrides?

Done

InBetweenNames:
Although, I don't condone `remove-la` in general, this makes sense to me.  Thanks!

thesamesam:
@InBetweenNames Could we consider turning it off in general using this flag? Upstream in Gentoo, we should strip it from ebuilds if it doesn't cause breakage.

I don't think a lot of users realise that portage-bashrc-mv does this by default (I didn't until recently!)

Issue #660, opened 2020-12-12
Use Flag Management Issues


WillPower3309:
Hello, when using `euse` or `flaggie`, two popular use flag management programs, both had issues with make.conf sourcing make.conf.lto with its relative file path, as well as make.conf.lto sourcing make.conf.lto.defines with its relative path. The fix was to change my make.conf and make.conf.lto to `source` using absolute file paths from root. Posting this in case the devs deem it worthy to modify things to be referenced by their absolute paths, as this is a potential issue that could arise for some users

telans:
What were the issues those programs were experiencing? It's a good idea, but I don't have issues myself

WillPower3309:
Both exhibited the error below:
```
Traceback (most recent call last):
  File "/usr/lib/python-exec/python3.8/flaggie", line 25, in <module>
    sys.exit(main(sys.argv))
  File "/usr/lib/python3.8/site-packages/flaggie/cli.py", line 192, in main
    mkconf = MakeConf(
  File "/usr/lib/python3.8/site-packages/flaggie/makeconf.py", line 537, in __init__
    self.parse(mf, path)
  File "/usr/lib/python3.8/site-packages/flaggie/makeconf.py", line 582, in parse
    self.files[fn] = self.MakeConfFile(fn, path)
  File "/usr/lib/python3.8/site-packages/flaggie/makeconf.py", line 455, in __init__
    f = codecs.open(path, 'r', 'utf8')
  File "/usr/lib/python3.8/codecs.py", line 905, in open
    file = builtins.open(filename, mode, buffering)
NotADirectoryError: [Errno 20] Not a directory: '/etc/portage/make.conf/make.conf.lto'
```

WillPower3309:
Its strange as the last line makes it appear as though it correctly interpreted the file path, but the issue disappears when I change it the the absolute file paths in my make.conf and make.conf.lto

jian-lin:
I have a issue here using `euse`.
```
$ sudo LANG=C euse -p www-client/firefox -E wayland
/etc/portage/make.conf: line 8: make.conf.lto: No such file or directory
Adding "www-client/firefox[wayland]" use flag in "/etc/portage/package.use/firefox"
```

The 8th line of `/etc/portage/make.conf` is `source make.conf.lto`.

jian-lin:
[...]

PR #696, opened 2021-02-01
change relative paths to absolute

WillPower3309:
This PR will allow for package and use flag managers to work with gentooLTO, referenced here: #320 and here: #660. In my opinion, this is the way to resolve both cleanly, although I understand any adversity towards absolute over relative paths. Seeing as portage configuration should be stored in `/etc/portage/` however, absolute paths shouldn't matter and I think this is the way to go about fixing those issues.


telans (approved):

InBetweenNames:
I agree!  In the future it would be nice to have a solution that works for cross-compilation and also Gentoo Prefix too.

PR #698, opened 2021-02-03
dev-db/firebird: package.cflags typo

AntarcticanWaffles:
*FLAGS=-flto* forcefully enables lto, and breaks all other *FLAGS. 
Fixed before, it seems


telans (approved):

InBetweenNames:
Thank you!  Good catch!

PR #706, opened 2021-02-28
dev-util/github-cli: disable LTO

Z5483:
log: https://pastebin.com/UNN2sX4U

The log doesn't seem to show any noticeable building error, but disabling LTO does make the package built successfully. I'm sorry for the lack of explanation!


mid-kid:
Can confirm that none of the other workarounds seem applicable here.

InBetweenNames:
Thanks!

PR #707, opened 2021-03-02
clang: disable LTO when DEPEND contains dev-qt

telans:
Applications that depend on dev-qt programs SIGABRT at runtime with LTO

usually with one of the below:
"QObject::connect: signal not found"
"QWidget: Cannot create a QWidget without QApplication"

rename 44-lto-clang-full.sh
re-add 44-lto-clang.sh symlink

### DIFF
--- a/sys-config/ltoize/files/bashrc.d/44-lto-clang-full.sh
+++ b/sys-config/ltoize/files/bashrc.d/44-lto-clang-full.sh
@@ -1,9 +1,0 @@
-LTO_CLANG_FULL() {
-    if [[ ${LTO_FULL} -eq 1 ]]; then
-        export CFLAGS="${CFLAGS//=thin}"
-        export CXXFLAGS="${CXXFLAGS//=thin}"
-        export LDFLAGS="${LDFLAGS//=thin}"
-    fi
-}
-
-BashrcdPhase configure LTO_CLANG_FULL
--- a/sys-config/ltoize/files/bashrc.d/44-lto-clang.sh
+++ b/sys-config/ltoize/files/bashrc.d/44-lto-clang.sh
@@ -1,0 +1,22 @@
+LTO_CLANG_FULL() {
+    if [[ ${LTO_FULL} -eq 1 ]]; then
+        export CFLAGS="${CFLAGS//=thin}"
+        export CXXFLAGS="${CXXFLAGS//=thin}"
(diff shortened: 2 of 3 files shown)


perfect7gentleman:
I've never had such problems.

telans:
https://github.com/InBetweenNames/gentooLTO/issues/444
https://bugreports.qt.io/browse/QTBUG-61710
https://bugreports.qt.io/browse/QTBUG-43556

perfect7gentleman:
cause I do it right.

telans:
> cause I do it right.

Yeah I'm 100% sure this is an issue (at least with znver2), as I just did a clean reinstall using clang from the start, with still the same issues.

barolo:
> > cause I do it right.
> 
> Yeah I'm 100% sure this is an issue (at least with znver2), as I just did a clean reinstall using clang from the start, with still the same issues.

It's not only znver2, I have an Intel platform to test and it's the same there

telans:
> General approach doesn't seem to be right as I have whole QT desktop built witrh LTO, plus dozen of apps and they work fine.

I'll do a clean install without LTO, snapshot and LTO everything Qt and see what happens. Might take a while though.

barolo:
> > General approach doesn't seem to be right as I have whole QT desktop built witrh LTO, plus dozen of apps and they work fine.
> 
> I'll do a clean install without LTO, snapshot and LTO everything Qt and see what happens. Might take a while though. What DE are you using?

KDE/Plasma

telans:
> General approach doesn't seem to be right as I have whole QT desktop built witrh LTO, plus dozen of apps and they work fine.

You mentioned in the other thread that you use GCC, not Clang. Does this still apply? I have these issues when building clang system-wide

barolo:
> > General approach doesn't seem to be right as I have whole QT desktop built witrh LTO, plus dozen of apps and they work fine.
> 
> You mentioned in the other thread that you use GCC, not Clang. Does this still apply? I have these issues when building with clang system-wide (not gcc)

I'm using GCC [11], last time I've checked it was the same with clang [will test again soon]

InBetweenNames:
So, should we leave these workarounds on a case-by-case instead of applying system wide?

barolo:
Definitelly case-by-case, some of the affected already disable lto themselves

telans:
Got it. I'll try narrow this down and add workarounds for affending programs then.

PR #721, opened 2021-03-22
sys-apps/systemd: set optimizations to O2

obi1:
with -O3 it compiles fine but homectl starts acting up.
homed logins and new user creations via `homectl create` fail with protocol error.


Uniminin:
Works fine on my end.

obi1:
I tried again with -O3 and with -O3 -falign-functions=32 and both give me `Failed to execute operation: Protocol error` when running `homectl create testuser`.

This is systemd-247.2-r4 with USE="acl cryptsetup gcrypt homed hwdb kmod lz4 lzma pam pcre policykit resolvconf seccomp (split-usr) sysv-utils zstd". What are your settings?

dr460nf1r3:
I can confirm that systemd needs O2 compile flags to successfully execute homectl create on my machine.

InBetweenNames:
Thanks!  We can revisit this later to figure out why `homectl` in particular is problematic.

PR #723, opened 2021-04-01
dev-qt/qtcore: Added Ofast workaround to fix scrolling in kde-apps/okular

asaparov:
Most forms of scrolling in okular have been broken for a while (mouse wheel over the document, arrow keys, page up/down, clicking on links in the document or in the thumbnails/outline). I recently found that it was caused by building qtcore with `-ffinite-math-only`.


InBetweenNames:
Thank you!

PR #732, opened 2021-04-23
net-libs/nodejs: disable LTO

leonardohn:
Passing the LTO flag through CFLAGS causes linking errors, however, LTO can still be enabled by a recently added useflag.


telans:
nodejs should already be filtering lto flags:
https://gitweb.gentoo.org/repo/gentoo.git/tree/net-libs/nodejs/nodejs-16.0.0-r1.ebuild#n106

leonardohn:
It seems that they fixed the issue in the meantime, so the workaround is not necessary anymore.

https://gitweb.gentoo.org/repo/gentoo.git/commit/net-libs/nodejs/nodejs-16.0.0-r1.ebuild?id=5f0fc09e53bfe231a41647ae29fd6bd495c4ed98

PR #737, opened 2021-04-30
media-libs/dav1d: Disable LTO

petronio:
Fails during linking with GCC 11.1.0


InBetweenNames:
Thank you!

lucius-martius:
Is the joint workarounds file (ltoworkarounds.conf) still supported? Isn't this supposed to go into lto.conf?

On my system this change has no effect since ltoworkarounds.conf is not symlinked to /etc/portage/package.cflags.

petronio:
You're correct, my mistake. I've had some other changes go into ltoworkarounds.conf as well, so I'll be moving them in another PR.

petronio:
https://github.com/InBetweenNames/gentooLTO/pull/751

PR #747, opened 2021-05-12
media-libs/mesa: disable IPAPTA to fix segfault

aruhier:
RADV (AMDGPU) and Intel GPUs crash with Vulkan and IPAPTA enabled.

Vulkan was not working since mesa 21.1 on my AMD 6800xt when I used RADV (segfault). I found this post advising to remove IPAPTA, which indeed fixed the issue for me: https://forums.gentoo.org/viewtopic-t-1134963-highlight-mesa.html


InBetweenNames:
Thank you!  Better to be safe with this one.

PR #749, opened 2021-05-18
dev-lang/erlang: Disable LTO

petronio:
Build failures starting with 24.0 . Not entirely surprising, as this release came with significant VM changes to support JIT.


InBetweenNames:
Thank you!

PR #751, opened 2021-05-23
Remove ltoworkarounds.conf and migrate mistaken commits

petronio:
I've mistakenly committed the dev-lang/erlang and media-libs/dav1d workarounds to the now unused ltoworkarounds.conf instead of the new lto.conf . Migrate the changes and delete the ltoworkarounds.conf file to avoid future confusion.


InBetweenNames:
Thank you!

PR #767, opened 2021-07-01
x11-base/xwayland: disable -fipa-pta to fix crash on GLX query

### LINKED ISSUE #766: x11-base/xwayland-21.1.1-r1 crashes with -fipa-pta under sway
First, some info:

```
GCC version: 11.1.0-r1
Glibc version: 2.33-r1
ACCEPT_KEYWORDS: "amd64 ~amd64"
```

**TL;DR: xwayland crashes with -fipa-pta (breaking the X connections of other applications) when querying GLX info**

After the latest round of updates, I encountered a strange issue: some XWayland applications would only work with specific flags (e.g. spotify with --disable-gpu) and certain XWayland applications would crash the whole embedded X server, resulting other applications not having a window anymore but still running (i.e. there still was music playing from Spotify etc.). This was the error they logged:
`X connection to :1 broken (explicit kill or server shutdown)`
The worst offender of crashing others was Lutris, and the other applications then logged that the X connection to the display was broken.

Turns out, this happened because Lutris queried for GLX information, and could be reproduced by simply running glxinfo in a terminal. After a lot of diagnosis, I figured out that XWayland crashed when trying to get display info. After trying various versions of Sway, Wlroots, and Mesa, I managed to figure out that building XWayland without -fipa-pta resolved the issue, which is why I suggest to add it to the LTOworkarounds.

B4ckslash:
XWayland applications querying for GLX information (glxinfo,
spotify, lutris, anything GPU accelerated) crash the X server
when XWayland is compiled with -fipa-pta.

Fixes #766

### DIFF
--- a/sys-config/ltoize/files/package.cflags/ipa-pta.conf
+++ b/sys-config/ltoize/files/package.cflags/ipa-pta.conf
@@ -23,4 +23,5 @@ net-p2p/monero *FLAGS-="${IPAPTA}" # ICE with -fipa-pta
 mail-client/thunderbird *FLAGS-="${IPAPTA}" # ICE with GCC 10.2.0 (seen with thunderbird 68.12.0)
 >=dev-lang/spidermonkey-78.3.1 *FLAGS-="${IPAPTA}" # Segfault during build with GCC 10.2.0
 >=media-libs/mesa-21.1.0 *FLAGS-="${IPAPTA}" # Segfault with vulkan
+x11-base/xwayland *FLAGS-="${IPAPTA}" # SIGABRT when querying for GLX information
 # END: -fipa-pta workarounds


2secslater:
I think you mean x11-base/xwayland as opposed to x11-apps/xwayland
https://packages.gentoo.org/packages/x11-base/xwayland

InBetweenNames:
Just waiting on the feedback to be addressed and then I'll merge this.

B4ckslash:
@2secslater you're right, thanks for the heads up! I somehow got it confused with `xisxwayland`, which is in x11-apps.

@InBetweenNames I amended the commit, it should be correct now.

InBetweenNames:
Thank you!

PR #776, opened 2021-08-09
dev-vcs/git: add -ffat-lto-objects workaround

CyB3RC0nN0R:
When compiling git without this workaround, cloning a remote repository results in a "bad object" error (the object in question seems to be random) most of the time.


BorisCarvajal:
Thanks for the workaround.

I found this recent GCC report that seems to be the same issue:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101949

PR #779, opened 2021-08-20
net-libs/nodejs: filter out -flto from *FLAGS; -Ofast workarounds

jelinekto:
net-libs/nodejs: There's `lto` USE flag that should be used to inject `-flto` instead, also LTO currently does not work with GCC 11 (you'll be asked to disable the USE flag when you try to emerge it).

app-editors/emacs: You'll be explicitly asked to disable `-ffinite-math-only` during emerge.

gui-apps/gammastep: Fork of `x11-misc/redshift`, `-ffinite-math-only` breaks it in the exactly same way.

### DIFF
--- a/sys-config/ltoize/files/package.cflags/lto.conf
+++ b/sys-config/ltoize/files/package.cflags/lto.conf
@@ -60,6 +60,7 @@ media-libs/mlt *FLAGS-=-flto*
 media-sound/pulseaudio *FLAGS-=-flto*
 media-video/ffmpeg *FLAGS-=-flto* #NOTE: Depending on your CPUFLAGS this may work with LTO.  The SSE intrinsics seem to cause problems in some cases. Only x86 requires workaround
 media-video/mplayer *FLAGS-=-flto*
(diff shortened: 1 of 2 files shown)


jelinekto:
I've added a (possibly) temporary workaround for Firefox; `-Ofast` and `-ffast-math` get filtered by flagomatic and specifying `-fno-honor-infinities -fno-honor-nans -fno-math-errno -ffinite-math-only -fassociative-math -freciprocal-math -fno-signed-zeros -fno-trapping-math -ffp-contract=fast` (aka the flags activated by Clang `-ffast-math`) causes the build to fail.

Obviously investigating Firefox build failures without a CPU that is beefy enough is an absolute bitch and would cause me to lose my will to live pretty fast, so I suggest falling back to `-fno-trapping-math -fno-math-errno` which makes it possible to at least build the package with both Clang and GCC for now. So far I can tell you removing `-ffinite-math-only` doesn't help.

InBetweenNames:
Thank you!

PR #774, opened 2021-07-27
media-libs/gst-plugins-good: disable LTO

aruhier:
Fails to compile.

Compiling with LTO gives the following error:
```
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: /var/tmp/portage/media-libs/gst-plugins-good-1.18.4/temp/cclqHELB.ltrans1.ltrans.o: in function `gst_deinterlace_method_yadif_init':
<artificial>:(.text+0x4c0c): undefined reference to `gst_yadif_filter_line_mode0_ssse3'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: <artificial>:(.text+0x4c13): undefined reference to `gst_yadif_filter_line_mode2_ssse3'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: <artificial>:(.text+0x4c73): undefined reference to `gst_yadif_filter_line_mode0_sse2'
/usr/lib/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../x86_64-pc-linux-gnu/bin/ld: <artificial>:(.text+0x4c7a): undefined reference to `gst_yadif_filter_line_mode2_sse2'
collect2: error: ld returned 1 exit status
```

I tried to disable other optimizations, such as IPA-PTA, forcing to use ld-gold, but it didn't solve the issue.

PR #791, opened 2021-10-09
sci-physics/root: disable LTO

jonesmz:
Can't compile these two packages without disabling lto.

There might be a less "hammer + nail" solution, but this change does let them compile and run.


jonesmz:
Removed gst-plugins-good in favor of: https://github.com/InBetweenNames/gentooLTO/pull/774

PR #793, opened 2021-10-16
dev-lang/python: Bump to 3.10 stable

aruhier:
Package-Manager: Portage-3.0.28, Repoman-3.0.3
Signed-off-by: Anthony Ruhier <aruhier@mailbox.org>


aruhier:
As now the official gentoo packages support LTO and PGO, it's probably not useful to maintain it here anymore.

PR #802, opened 2021-11-03
dev-vcs/git: disable -fipa-pta

bluecube:
Added a workaround for GCC 11.2.0 bug that causes git to be miscompiled with `-flto -fipa-pta`.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=101949
https://bugs.gentoo.org/809419


InBetweenNames:
Nice, looks like it was fixed upstream, we can remove this on the next point release.

PR #823, opened 2022-01-17
sys-libs/efivar-38: disable LTO

aruhier:
Segfaults during linking.


yibaba:
It works for me with `sys-libs/efivar *FLAGS-=-flto*`

wolfwood:
I also had to make this change, locally. works for me

PR #824, opened 2022-01-18
app-misc/openrgb: segmentation fault will occur on launch with LTO enabled

x86k:
app-misc/openrgb will complete build successfully however will generate a core dump due to a segmentation fault, on launch.


InBetweenNames:
Thank you!

PR #825, opened 2022-01-18
kde-apps/cantor: will not build with -O3

x86k:
kde-apps/cantor will fail build with -O3 but works perfectly fine with -O2, graphite, and other optimizations enabled by ltoize.


InBetweenNames:
Thank you!

PR #833, opened 2022-03-19
dev-db/postgresql: disable lto

gosuloli:
I was having an issue compiling dev-db/postgresql with lto. 
Disabling lto seems to fix the issue.


JustArchi:
What issues did you have? I compiled `dev-db/postgresql-14.2-r1` just fine with packages currently available in Gentoo testing for amd64.

Issue #808, opened 2021-12-03
webkit-gtk x86_64-pc-linux-gnu-g++: error: unrecognized command-line option ‘-stdlib=libstdc++’


Jacekalex:
My OS:
`Linux version 5.10.83-g1 (root@localhost) (gcc (Gentoo Hardened 11.2.0 p1) 11.2.0, GNU ld (Gentoo 2.37_p1 p0) 2.37) #1 SMP PREEMPT Wed Dec 1 23:08:57 CET 2021`

clang.conf options:
` * /etc/portage/package.cflags/clang.conf -> net-libs/webkit-gtk: CXXFLAGS+=-stdlib=libstdc++ # error: use of undeclared identifier 'LC_ALL'`

is incompatible with GCC-11.2.0.

Error:

```
    Change Dir: /var/tmp/portage/net-libs/webkit-gtk-2.32.4/work/webkit-gtk-2.32.4_build/CMakeFiles/CMakeTmp
    
    Run Build Command(s):/usr/bin/ninja cmTC_b0d8e && [1/2] Building CXX object CMakeFiles/cmTC_b0d8e.dir/testCXXCompiler.cxx.o
    FAILED: CMakeFiles/cmTC_b0d8e.dir/testCXXCompiler.cxx.o 
    /usr/bin/x86_64-pc-linux-gnu-g++   -march=native -O2 -pipe -DNDEBUG -march=native -O2 -pipe -stdlib=libstdc++ -Wl,-O1 -Wl,--as-needed -o CMakeFiles/cmTC_b0d8e.dir/testCXXCompiler.cxx.o -c testCXXCompiler.cxx
    x86_64-pc-linux-gnu-g++: error: unrecognized command-line option ‘-stdlib=libstdc++’
    ninja: build stopped: subcommand failed.
```

Cheers

TheMBL09898:
Months later, same here... 
**emerge --info**

```

Portage 3.0.30 (python 3.9.10-final-0, default/linux/amd64/17.1/desktop, gcc-11.2.1, glibc-2.34-r10, 5.16.11-gentoo-dist x86_64)
=================================================================
System uname: Linux-5.16.11-gentoo-dist-x86_64-Intel-R-_Core-TM-_i7-8750H_CPU_@_2.20GHz-with-glibc2.34
KiB Mem:    16105744 total,  13334524 free
KiB Swap:    4194296 total,   4194296 free
[...]

PR #836, opened 2022-03-29
net-libs/webkit-gtk: Fix compilation error caused by -stdlib=libstdc++

Skyblueborb:
Title: net-libs/webkit-gtk: Fix compilation error caused by -stdlib=libstdc++

net-libs/webkit-gtk doesn't pass the tests set by g++, therefore compilation fails.
[Here's](https://github.com/InBetweenNames/gentooLTO/issues/808) the GitHub issue.


TheMBL09898:
You commented the line so that means the error has been fixed? I don't understand.

Skyblueborb:
Yup, that line was causing the error and I didn't feel like removing it

jpgrego:
Hey, I had the same issue and successfully applied the same solution. I was wondering if there was some reason why this wasn't merged?

TheMBL09898:
I'm using gcc, this is part of clang.conf... Maybe this fixes something for clang users?

PR #856, opened 2022-09-11
dev-lang/ldc2::dlang: required -ffat-lto-objects

mccarverone:
dev-lang/ldc2 is from dlang overlay. Requires -ffat-lto-objects to compile, otherwise it fails with lots of unrecognized symbols.

This is a compiler so it definitely requires further tests to ensure that everything works fine.

[build.log](https://github.com/InBetweenNames/gentooLTO/files/9543469/build.log)


Connor-GH:
Tested with 1.35. Works with clang ``-flto`` but not with gcc. Still needs ``-ffat-lto-objects``.

Connor-GH:
Considering I build and bootstrap ldc2 with itself and also build dmd with it and many personal projects, I would consider ldc2 fine to use with clang lto.

PR #866, opened 2023-01-05
dev-lang/duktape: disable Ofast

blackteahamburger:
Can't compile with macro \_\_FAST_MATH\_\_.


Xpeq7 (approved):
I suggest changing comment to:
(same as before) `# preprocessor in the source throws an error if __FAST_MATH__ macro is defined`

blackteahamburger:
Close since this project is in maintenance mode.

PR #867, opened 2023-01-05
sys-apps/systemd: disable Ofast

blackteahamburger:
Configure fails with error.


blackteahamburger:
I found that the O3 workaround for systemd is only required if the `homed` USE flag is enabled according to the comment. So is there a way to detect that USE flag and decide if a workaround is needed?

blackteahamburger:
Only #872?

blackteahamburger:
Close this since this project is in maintenance mode.

PR #868, opened 2023-01-06
sys-auth/polkit: diable Ofast

blackteahamburger:
Build fails with error.


blackteahamburger:
Close since this project is in maintenance mode.

PR #870, opened 2023-01-06
media-libs/rnnoise: disable Ofast

blackteahamburger:
Preprocessor in the source throws an error if ffast-math is enabled.


blackteahamburger:
Compiles just fine by just using `-DFLOAT_APPROX`

